PR #267, opened 2014-09-27
Ajax paginate

carlqt:
Just added a AJAX functionality to issues paginations


jroes:
Seems like this languished, build's not passing, and no one has really stepped in to make a case for ajax here. Gonna close for now, feel free to re-open if anyone wants to engage in heated debate! :)

PR #289, opened 2014-11-29
Fix mail sent bug for repo without issues

igas:
When user have subscriptions and add new repo without issues the mail anyway sent with wrong link with issue number from different repo.


igas:
@schneems I just rebased it to fix conflicts. Can you take a look?

schneems:
Great catch, I see what was happening. Thanks :heart:

PR #305, opened 2015-01-14
Hide callbacks from public API

igas:
Also change `downcase_name` from `downcase` to `downcase!` to match `strip_whitespaces`.


prathamesh-sonpatki:
@igas Thank you :yellow_heart:

Issue #307, opened 2015-01-16
Error when click in "Send new Issue"


duduribeiro:
I'm trying to click in "Send new Issue" button in http://www.codetriage.com, but I get a "We're sorry, but something went wrong"
![click_button](http://i.imgur.com/QSYmntY.png)
![error](http://i.imgur.com/weviO75.png)

When running local, the IssueAssignmentsController is trying to call a undefined_method
![error_desc](http://i.imgur.com/eXIDR9L.png)

prathamesh-sonpatki:
Closed by #308

PR #308, opened 2015-01-16
Fix the undefined method call. Calling the resque method

duduribeiro:
Fixing the issue #307 .


prathamesh-sonpatki:
Looks good! Thanks :smile:

Issue #304, opened 2015-01-13
Exclude own issues and pull requests from triage lists where no replies


RupW:
I've just signed up to triage a new repository, https://github.com/jaredholdcroft/postmark-java, and the issue it sent me to look at was this one:
- https://github.com/jaredholdcroft/postmark-java/pull/12 - Switch to using the HTTPS endpoint

which is a pull request I've just raised myself.

There isn't a lot I can do with that one! Maybe exclude my own issues and pull requests, or at least the ones without any activity from other users, from the set of things you sent me?

schneems:
We should be doing this already, we check an issue to see if you're the owner or already commented before sending out. Must be a bug somewhere

—

bhollan:
I had the same thing.  I got the same issue two days in a row after commenting on it more than once the day before.

mario-d-s:
@RupW, @schneems Has this bug been squished? Maybe the issue can be closed since there has been no activity for almost three years.

RupW:
I certainly haven't seen it since, but I haven't been raising a lot of issues or PRs either :-/

schneems:
Closing for now. Thanks

mario-d-s:
@schneems I'm afraid the bug is still here and I have the proof:

![CodeTriage email](https://user-images.githubusercontent.com/4290238/39351991-99998b46-4a03-11e8-93b9-78181b20957e.png)

#706 is an issue I raised myself.

What strikes me is that I raised that issue less than 24h ago, as was the case with @RupW's pull request. Perhaps this gives a hint to the root cause of this problem?

PR #311, opened 2015-01-29
Excludes own issues and PRs from triage list

O-I:
Hi.

This pull request attempts to fix issue #304. I'm not sure if this is the best way to do this (or even if this is the desired behavior), so please review at your convenience and let me know how you'd like it changed.

Before these commits, the behavior was to exclude the following from a given user's triage list:
1. issues (and pull requests) the user has commented on
2. closed issues
3. pull requests when the user's `skip_issues_with_pr?` flag is set to `true`

However, a user's triage list would include one's own uncommented issues (and PRs if `skip_issues_with_pr?` is `false`). For example, if I subscribe to `codetriage/codetriage` and submit a pull request to which no one else responds, I could very well end up getting an email asking me to triage my own issue.

When an issue is created in the database, the user who created the issue isn't being stored anywhere. And the comments don't include any information on the user who originally filed the issue. So, I added a `creator` attribute to the `issues` table that stores the Github username of the person who filed the issue and used that to exclude one's own issues from a user's triage list.

Assuming this is an acceptable implementation, is this the behavior users want? I think it's reasonable to not want to see your own issues and PRs in your triage list, but I don't know — maybe it makes sense to some to have a `skip_my_own_issues?` setting.

Let me know your thoughts when you have time and I'd be happy to fix my fix however you'd like.

Cheers,
O-I

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -51,7 +51,7 @@ class Issue < ActiveRecord::Base
   end
 
   def created_by?(user)
-    user.github == self.creator
+    user.github == self.created_by
   end
 
   def repo_name
@@ -83,7 +83,7 @@ class Issue < ActiveRecord::Base
   def update_from_github_hash!(issue_hash)
     self.update_attributes(title:           issue_hash['title'],
                            url:             issue_hash['url'],
-                           creator:         issue_hash['user']['login'],
+                           created_by:      issue_hash['user']['login'],
                            last_touched_at: DateTime.parse(issue_hash['updated_at']),
                            state:           issue_hash['state'],
                            html_url:        issue_hash['html_url'],
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -16,8 +16,9 @@ class Issue < ActiveRecord::Base
       update_issue!
       return false if commenting_users.include?(user.github)
     end
-    return false  if closed?
-    return false  if pr_attached? && user.skip_issues_with_pr?
+    return false if closed?
+    return false if created_by?(user)
+    return false if pr_attached? && user.skip_issues_with_pr?
     true
   end
 
@@ -49,6 +50,10 @@ class Issue < ActiveRecord::Base
     state == OPEN
   end
 
+  def created_by?(user)
+    user.github == self.creator
+  end
+
   def repo_name
     self.repo.name
   end
@@ -78,6 +83,7 @@ class Issue < ActiveRecord::Base
   def update_from_github_hash!(issue_hash)
     self.update_attributes(title:           issue_hash['title'],
                            url:             issue_hash['url'],
+                           creator:         issue_hash['user']['login'],
                            last_touched_at: DateTime.parse(issue_hash['updated_at']),
                            state:           issue_hash['state'],
                            html_url:        issue_hash['html_url'],
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -17,7 +17,7 @@ class Issue < ActiveRecord::Base
       return false if commenting_users.include?(user.github)
     end
     return false if closed?
-    return false if created_by?(user)
+    return false if created_by?(user) && user.skip_my_own_issues_and_prs?
     return false if pr_attached? && user.skip_issues_with_pr?
     true
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150211163514) do
+ActiveRecord::Schema.define(version: 20150211235706) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -69,12 +69,12 @@ ActiveRecord::Schema.define(version: 20150211163514) do
   end
 
   create_table "users", force: :cascade do |t|
-    t.string   "email",                  default: "",                                   null: false
-    t.string   "encrypted_password",     default: "",                                   null: false
+    t.string   "email",                      default: "",                                   null: false
+    t.string   "encrypted_password",         default: "",                                   null: false
     t.string   "reset_password_token"
     t.datetime "reset_password_sent_at"
     t.datetime "remember_created_at"
-    t.integer  "sign_in_count",          default: 0
+    t.integer  "sign_in_count",              default: 0
     t.datetime "current_sign_in_at"
     t.datetime "last_sign_in_at"
     t.string   "current_sign_in_ip"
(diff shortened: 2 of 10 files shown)


prathamesh-sonpatki:
@O-I This looks good. I just think rather than `creator`, we can say `created_by`. Also we should provide setting for skip_my_own_issues_and_prs in the UI.

O-I:
@prathamesh-sonpatki You're right — `created_by` is a better name than `created` here. :smiley: 

As for adding a `skip_my_own_issues_and_prs` flag, do you want the default to be `true` or `false`? I think `true` makes more sense to me as a default, but I can see an argument for keeping the behavior of the app the same now that users will have a choice.

Let me know and I'll make the necessary changes.

prathamesh-sonpatki:
@O-I I think keeping the flag `false` by default makes sense. This is just to keep the original behavior intact.

parndt:
Yeah, I quite like the behaviour of seeing my own issues!

db/schema.rb
```diff
@@ -11,15 +11,15 @@
...
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "issue_assignments", force: true do |t|
+  create_table "issue_assignments", force: :cascade do |t|
     t.integer  "issue_id"
-    t.datetime "created_at",                           null: false
-    t.datetime "updated_at",                           null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
```
parndt: how come this changed the DB so much?

prathamesh-sonpatki: I think some changes due to Rails 4.2.

parndt: Good times

O-I: I agree we should take a closer look at what's going on with the schema. It looks like some things were manually added to this file in these two commits: [1](https://github.com/codetriage/codetriage/commit/41a4b35f5252b0649a1105f962b772162e076f43) & [2](https://github.com/codetriage/codetriage/commit/9dd20530ab3b3a1176f643a658e93ceb110bdb2a).

That means when I ran my migration, it clobbered all of these edits. I don't think explicitly adding `null: false` on timestamps is necessary anymore after this [commit](https://github.com/rails/rails/commit/a939506f297b667291480f26fa32a373a18ae06a) into Rails, but I'll add back if you want. More below.

db/schema.rb
```diff
@@ -27,94 +27,73 @@
...
-  create_table "opro_client_apps", force: true do |t|
-    t.string   "name",        limit: nil
-    t.string   "app_id",      limit: nil
-    t.string   "app_secret",  limit: nil
-    t.text     "permissions"
-    t.integer  "user_id"
-    t.datetime "created_at",              null: false
-    t.datetime "updated_at",              null: false
-  end
-
```
O-I: These two tables `opro_auth_grants` and `opro_client_apps` were added in the first commit referenced above. They don't seem to have corresponding migrations. I'll add them back in manually.

This is just for my edification, but isn't changing the schema directly like this generally not a good idea? I'm happy to restore it, but won't we have this issue again any time anyone needs to make a migration? I'm just trying to understand why these changes seem to have been made directly. Does it have anything to do with these tables corresponding to POROs?

parndt: I don't think we should be adding things back in manually, I was just curious :)

O-I: I'm curious, too. :smiley: I guess I'll just continue with what I have now for `schema.rb`.

app/models/issue.rb
```diff
@@ -16,8 +16,9 @@ def valid_for_user?(user, skip_update = Rails.env.test?)
       update_issue!
       return false if commenting_users.include?(user.github)
     end
-    return false  if closed?
-    return false  if pr_attached? && user.skip_issues_with_pr?
+    return false if closed?
+    return false if created_by?(user) && user.skip_my_own_issues_and_prs?
+    return false if pr_attached? && user.skip_issues_with_pr?
     true
```
prathamesh-sonpatki: @O-I All the `return false` statements can be be extracted into a separate method.

O-I:
@prathamesh-sonpatki Before I push, this is how I have it locally:

```ruby
def valid_for_user?(user, skip_update = Rails.env.test?)
    unless skip_update
      update_issue!
      return false if commenting_users.include?(user.github)
    end
    valid_for?(user)
  end

# snip

private

# snip

def valid_for?(user)
    !closed? && !(created_by?(user) && user.skip_my_own_issues_and_prs?) &&
                !(pr_attached?      && user.skip_issues_with_pr?)
end
```

Do you have a suggestion for a better name than `valid_for?`? Also, is that boolean expression too confusing? Is this clearer?

```ruby
def valid_for?(user)
  return false if closed?
  return false if created_by?(user) && user.skip_my_own_issues_and_prs?
  return false if pr_attached? && user.skip_issues_with_pr?
  true
end
```

prathamesh-sonpatki:
@O-I I think it can be something like this

```ruby
def valid_for_user?(user, skip_update = Rails.env.test?)
   update_issue!  unless skip_update
   still_valid_for?(user)
end

...

def still_valid_for?(user)
  return false if closed?
  return false if created_by?(user) && user.skip_my_own_issues_and_prs?
  return false if pr_attached? && user.skip_issues_with_pr?
  return false if commenting_users.include?(user.github)
  true
end

```

What do you think?

O-I:
@prathamesh-sonpatki I'd be really happy if we could get `valid_for_user?` to be what you have. If I take `return false if commenting_users.include?(user.github)` out of the `unless` statement, though, I get 4 `VCR::Errors::UnhandledHTTPRequestError` that look like:

```
  1) Error:
ReposControllerTest#test_do_not_send_email_for_repo_without_issues:
VCR::Errors::UnhandledHTTPRequestError: 

================================================================================
An HTTP request has been made that VCR does not know how to handle:
  GET https://api.github.com/repos/refinery/refinerycms/issues/2272/comments

VCR is currently using the following cassette:
  - /Users/Raevynheart/Code/forks/codetriage/test/vcr_cassettes/create_repo_refinery.yml
  - :record => :once
  - :match_requests_on => [:method, :uri]

Under the current configuration VCR can not find a suitable HTTP interaction
to replay and is prevented from recording new requests. There are a few ways
you can deal with this:

  * If you're surprised VCR is raising this error
    and want insight about how VCR attempted to handle the request,
    you can use the debug_logger configuration option to log more details [1].
  * You can use the :new_episodes record mode to allow VCR to
    record this new request to the existing cassette [2].
  * If you want VCR to ignore this request (and others like it), you can
    set an `ignore_request` callback [3].
  * The current record mode (:once) does not allow new requests to be recorded
    to a previously recorded cassette. You can delete the cassette file and re-run
    your tests to allow the cassette to be recorded with this request [4].

[1] https://www.relishapp.com/vcr/vcr/v/2-5-0/docs/configuration/debug-logging
[2] https://www.relishapp.com/vcr/vcr/v/2-5-0/docs/record-modes/new-episodes
[3] https://www.relishapp.com/vcr/vcr/v/2-5-0/docs/configuration/ignore-request
[4] https://www.relishapp.com/vcr/vcr/v/2-5-0/docs/record-modes/once
================================================================================


    app/models/issue.rb:72:in `commenting_users'
    app/models/issue.rb:118:in `still_valid_for?'
    app/models/issue.rb:20:in `valid_for_user?'
    app/models/issue_assigner.rb:44:in `assign_issue_for_sub'
    app/models/issue_assigner.rb:13:in `block (2 levels) in assign'
    app/models/issue_assigner.rb:12:in `times'
    app/models/issue_assigner.rb:12:in `each'
    app/models/issue_assigner.rb:12:in `map'
    app/models/issue_assigner.rb:12:in `block in assign'
    app/models/issue_assigner.rb:11:in `each'
    app/models/issue_assigner.rb:11:in `assign'
    app/models/repo_subscription.rb:20:in `block in <class:RepoSubscription>'
    app/controllers/repos_controller.rb:46:in `create'
    test/functional/repos_controller_test.rb:26:in `block (2 levels) in <class:ReposControllerTest>'
    test/functional/repos_controller_test.rb:25:in `block in <class:ReposControllerTest>'
```

Any advice on that?

prathamesh-sonpatki:
@O-I I checked it locally. We are pulling the issues in `after_create` callback after creating repo. I am not sure why its not getting recorded in vcr cassets.

In your stack trace also, the issue is getting pulled dynamically

`https://api.github.com/repos/refinery/refinerycms/issues/2272/comments`

The issue with 2272 is not present in fixtures. Need to debug more :smile:

nateberkopec:
Stale, feel free to reopen.

Issue #312, opened 2015-02-11
Subject line in daily triage email incorrectly pluralizes "issue"


iKevinY:
I configured my account to only send one issue per day, but the subject line reads "Help Triage 1 Open Source Issues". Ruby is completely out of my depth, but could this be due to `@assignments.count` not returning the correct value [here](https://github.com/codetriage/codetriage/blob/master/app/mailers/user_mailer.rb#L13) in `UserMailer`? (Would changing `.count` to `.size` fix the issue?)

![CodeTriage Subject Line](https://cloud.githubusercontent.com/assets/2434728/6142444/6e6d0d8c-b16b-11e4-971e-a95ed27fab01.png)

marceloboeira:
@iKevinY I guess that is not the case, `.count` is used because `@assigments` can be nil/empty I guess, and `size` is only for array instances. I would say to change it anyway, because you need to trust your own inputs, now coming from front-end/client ...

O-I:
@iKevinY Interesting. I tried to reproduce this by setting my max issue per day count to 1. I just got my triage email and it displays "Help Triage 1 Open Source Issue" correctly. Can you pop in and let us know if the problem persists in your next email?

iKevinY:
@O-I Sure. The email I received shortly before your comment still read "1 Open Source Issues" – I'll comment again after getting tomorrow's email.

iKevinY:
The email I received today also reads "1 Open Source Issues".

O-I:
@iKevinY OK. Thanks for letting me know. I'm pretty sure what you suggest will fix this, but looking into this [commit](https://github.com/codetriage/codetriage/commit/56ef475d1ea9a847ea2261a3faf9c53fd2dfddb8) and associated Rails [issue](https://github.com/rails/rails/issues/16171) to understand why.

iKevinY:
@O-I Want me to open a PR making that change?

O-I:
Go for it. It looks like an innocent enough fix.

iKevinY:
Looks like #313 did the trick! My latest triage email correctly reads "1 Open Source Issue".

schneems:
Great! Thanks.



—

PR #313, opened 2015-02-13
Determine plurality of "issue" using .size method

iKevinY:
Changes `UserMailer.send_daily_triage()` to use `@assignments.size` as the argument to `"Issue".pluralize()` rather than `@assignments.count`. I'm not too familiar with Ruby so I don't know if this will actually fix the underlying issue, but I presume that since `@assignments.size` is returning the correct number of issues, then the behaviour will either become fixed or stay the same. Addresses #312.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -10,7 +10,7 @@ class UserMailer < ActionMailer::Base
     subject = ""
     @days   = @user.days_since_last_clicked
     subject << "[#{@days} days] " if @days > @max_days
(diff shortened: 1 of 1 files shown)


schneems:
:+1: worth a shot. The two do slightly different things, but they should be returning the same result. :shrug:

iKevinY:
@schneems Do commits to `master` get pushed to production immediately? If so, I'll post a follow-up in #312 tomorrow with an update on whether this has actually fixed the issue.

schneems:
Production is up to date with master as long as tests are passing. Everything in master is currently running on production.

iKevinY:
Ah, I see.

PR #321, opened 2015-03-21
 Bump to Rails 4.2.1

prathamesh-sonpatki:


prathamesh-sonpatki:
//cc @schneems @arthurnn

arthurnn:
:tada: ! :ship:

schneems:
It looks like this change dropped our RAM useage dramatically, i don't know why.

## Before


## After


Notice we're not swapping anymore

prathamesh-sonpatki:
Could this be related to our switch to Ruby 2.2.1?

schneems:
Maybe, however Ruby 2.2.1 was deployed a LONG time ago, I took that "before" screenshot pretty recently.  Maybe it's the combination of 2.2.1 and Rails 4.2.1 or something. :shrug:

schneems:
Looks like it's 2 things, I was playing with GC settings and i set `RUBY_GC_HEAP_GROWTH_FACTOR=1.1` which when reverted brings back the memory errors. I had this set before and we were throwing errors, so It's probably the combination of this and Rails 4.2.1.

I may try to revert to reproduce the memory issues to confirm. While i'm glad the issue is fixed, I wish I knew why it happened in the first place or nothing is stopping the issue from mysteriously coming back.

schneems:
If I rollback to v389 which corresponds to this commit: https://github.com/codetriage/codetriage/commit/6a13453 the memory problems come back. 

If I rollback to v390 which corresponds to https://github.com/codetriage/codetriage/commit/13f3f1e then the memory problems go away. So something definitely helped bring down memory when we went from the RC to the release, not sure what it was. I cannot reproduce the memory growth locally, only on production. I'm going to look into heap dumping sometime this week:

schneems:
Here's the diff between the RC and official release https://github.com/rails/rails/compare/v4.2.1.rc3...v4.2.1

I don't see anything that would help with memory. Maybe it's due to another dependent gem being updated.

PR #323, opened 2015-03-29
Refactor layouts

igas:
I removed some unnecessary `id` and `class`es and unify layout, we don't need 2 different layouts.


schneems:
Thanks so much, this is awesome!

I verified locally, everything looks the same (refactor successful!). Thanks :heart: :heart: :heart:

PR #322, opened 2015-03-28
Remove missing closing tag

igas:

PR #330, opened 2015-04-20
Change erb to slim

igas:
I started move from erb to slim as discussed https://github.com/codetriage/codetriage/pull/322#issuecomment-88970932
What do you think @schneems? If you'll like it, I'll finish the rest.
/cc @prathamesh-sonpatki


schneems:
I think it looks good, do you have any feelings @prathamesh-sonpatki ?

prathamesh-sonpatki:
:+1: from me :smile:

PR #331, opened 2015-04-21
Remove unused methods from repo model and add test for Repo#api_issues_path

gazay:
@igas pointed me on those unused methods. I grep all project's folders and remove them as they are unused. Also I've removed tests for those methods and replace them with test for used `api_issues_path` method.


prathamesh-sonpatki:
:+1: Thanks @gazay @igas :smile:

PR #332, opened 2015-04-21
Finish erb to slim transition

igas:
Together with @gazay we just finished erb to slim transition. We tried to keep everything as is, and we have plans to fix view issues we found in near future.


schneems:
Looks good to me.

Issue #314, opened 2015-02-21
Provide helpful errors for invalid input when adding new repos


O-I:
I think this is a separate issue from #241, but feel free to close if you think it's a duplicate or not worth worrying about at this stage.

Currently, entering any sort of invalid input into any of the form fields on `repos/new` and submitting throws up a 422 or 500 error. This includes scenarios like accidentally clicking 'Add Repo' before one or more form fields are completed, pasting in a malformed link, and even trying to add a repo you already triage.

Ideally, it would be nice if, instead of the app blowing up, the page was re-rendered with helpful errors like "those fields can't be blank" or "this repo doesn't seem to exist; you may have mistyped something" or "you're already triaging this repo".

Most of these scenarios are pretty unlikely so it's not that big of a deal, but I'm all for more foolproof forms if it all possible.

chrisccerami:
Hey I took a shot at submitting a pull request for this issue, #333. It's my first real open source contribution, so let me know if anything needs to be cleaned up. I'm not sure if the previous functionality was intentional, but considering the controller does have code to re-render the page on failure, I assume that's the intended result.

schneems:
Closed via #333. It's merged in, should auto-deploy in a few minutes. Give it a whirl maybe in 30 min and let me know if that works for you, if not i'll re-open issue. Thanks for the report and the fix :heart:

PR #333, opened 2015-05-01
Fix exception when submitting invalid repo params #314

chrisccerami:
Rather than raising an exception when the user enters invalid params to find a repo, it should now render the form with appropriate errors.


prathamesh-sonpatki:
@chrisccerami What was the error that we were getting with invalid params?

chrisccerami:
If a user click "Add Repo" without filling in the form, the site would crash. The same would happen if the user entered something that wasn't a real repo, like if they had a typo in their entry. Now it correctly renders the /repos/new page again with error messages.

app/controllers/repos_controller.rb
```diff
@@ -85,8 +83,12 @@ def parse_params_for_repo_info
       if params[:url]
         params[:url].match(/^https:\/\/github\.com\/([^\/]*)\/([^\/]*)\/?$/)
         params[:repo] ||= {}
-        params[:repo][:user_name] = $1
-        params[:repo][:name] = $2
+        params[:repo][:user_name] = $1.to_s
+        params[:repo][:name] = $2.to_s
```
schneems: Didn't realize this match logic was in here. We don't need to do it now, but we should eventually switch to a match block and named matches

```ruby
params[:repo] ||= {}
params[:url].match(/^https:\/\/github\.com\/(?<user_name>[^\/]*)\/(?<name>[^\/]*)\/?$/) do |m|
  params[:repo][:user_name] = m[:user_name]
  params[:repo][:name]      = m[:name]
end
```

app/controllers/repos_controller.rb
```diff
@@ -85,8 +83,12 @@ def parse_params_for_repo_info
...
       end
     end
+
+    def params_blank?
+      repo_params.values.any? { |param| param.blank? }
```
schneems: You can shorten this up

```ruby
repo_params.values.any?(&:blank?)
```

app/views/shared/_add_repos_list.html.slim
```diff
@@ -1,4 +1,4 @@
-- if repos.size == 0
+- if repos.blank? || repos.size == 0
```
schneems: The `blank?` method will cover when an array is empty.

```
[].blank?
=> true
```

We could remove the `repos.size == 0 ` check.

schneems:
I'm a big fan of nice helpful error messages. Thanks! :heart: I made some style comments that aren't crucial to execution but would be nice to have. If you're interested in another PR, could you clean up some of the items here? Welcome to the codetriage contributor list :rocket:

chrisccerami:
Yeah I can clean up those comments in a few hours. Thanks for the feedback.

chrisccerami:
Ok I fixed the last two comments. I also realized the site crashes if you try to submit a repo that's already been used. Let me know if you'd like me to address that in this PR or if I should open another for it.

schneems:
Can you open a new PR for the new issue please? Thanks for your help

schneems:
This is still failing ` undefined method `errors' for nil:NilClass`

PR #334, opened 2015-05-01
Reduce queries when rendering new repo page

chrisccerami:
When rendering the new repos page, Repo#exists_with_name? was querying the database for each repo to see if it's already been added. I replaced these with a single query so each repo could be checked without going back to the database. I'm not sure what the performance difference will be on production, but on local I noticed a slight increase in speed. With multiple queries I was noticing a typical response of ~950ms to load the page, and with a single query it averaged ~750ms. This was when testing with my own GitHub account with just under 40 repos, so the performance gains will probably vary for users with more or less repos on their account.

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -94,7 +94,7 @@ class Repo < ActiveRecord::Base
 
   def self.exists_with_name?(name, repos)
     user_name, repo_name = name.downcase.split(?/)
-    repos.map { |r| r.full_name }.include?(name)
+    repos.map { |r| r.full_name }.include?("#{user_name}/#{repo_name}")
   end
 
   def self.order_by_subscribers
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -94,7 +94,7 @@ class Repo < ActiveRecord::Base
 
   def self.exists_with_name?(name, repos)
     user_name, repo_name = name.downcase.split(?/)
-    repos.map { |r| r.name }.include?(name)
+    repos.map { |r| r.full_name }.include?(name)
   end
 
   def self.order_by_subscribers
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -92,9 +92,9 @@ class Repo < ActiveRecord::Base
     background_populate_issues(self.id)
   end
 
-  def self.exists_with_name?(name)
+  def self.exists_with_name?(name, repos)
     user_name, repo_name = name.downcase.split(?/)
-    Repo.exists?(user_name: user_name, name: repo_name)
+    repos.map { |r| r.name }.include?(name)
   end
 
   def self.order_by_subscribers
--- a/app/views/repos/new.html.slim
+++ b/app/views/repos/new.html.slim
@@ -18,13 +18,13 @@
 
   .tab-content
     .tab-pane.active#own
-      = render 'shared/add_repos_list', repos: @own_repos
+      = render 'shared/add_repos_list', repos: @own_repos, all_repos: Repo.all
 
     .tab-pane.active#starred
-      = render 'shared/add_repos_list', repos: @starred_repos
+      = render 'shared/add_repos_list', repos: @starred_repos, all_repos: Repo.all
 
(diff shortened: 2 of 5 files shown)


app/views/shared/_add_repos_list.html.slim
```diff
@@ -15,7 +15,7 @@
             td= repo["full_name"]
             td #{repo["open_issues_count"]} issues
             td
-              - if Repo.exists_with_name?(repo["full_name"])
+              - if Repo.exists_with_name?(repo["full_name"], current_user.repos)
                 p Added!
```
prathamesh-sonpatki: We can't pass current user's repos here, because some other user might have added the same repo already. I think original logic of checking with all repos in the table was correct.

app/models/repo.rb
```diff
@@ -92,9 +92,9 @@ def populate_issues!
     background_populate_issues(self.id)
   end
 
-  def self.exists_with_name?(name)
+  def self.exists_with_name?(name, repos)
     user_name, repo_name = name.downcase.split(?/)
-    Repo.exists?(user_name: user_name, name: repo_name)
+    repos.map { |r| r.full_name }.include?(name)
```
prathamesh-sonpatki: Here the user name is github repo's username. For eg. for schneems/codetrigage, user name is schneems and repo name is codetrigage.

prathamesh-sonpatki:
@chrisccerami Thanks for the patch but I think original logic is correct. To improve the speed, we can added indexes for `user_name` and `name` columns of repos table. Can you add those indexes?

chrisccerami:
Sure thing. I'll update the PR tonight when I get home.

chrisccerami:
Ok, so I fixed the exists_with_name method to use the correct format for the repo name, and I changed it to used all of the Repos rather than just the current_user's repos. This way the functionality of exists_with_name doesn't change, but it still reduces the number of database queries to just one, and reduces the load time by ~150ms. I also added the indices to the repos table. I realize you may still want the method to go back to its original implementation, and that's fine, but I feel like making a separate query for each repo is overkill.

app/views/repos/new.html.slim
```diff
@@ -18,13 +18,13 @@
 
   .tab-content
     .tab-pane.active#own
-      = render 'shared/add_repos_list', repos: @own_repos
+      = render 'shared/add_repos_list', repos: @own_repos, all_repos: Repo.all
```
prathamesh-sonpatki: I think we can skip passing all repos. We can calculated them in the model method itself if needed.

app/models/repo.rb
```diff
@@ -92,9 +92,9 @@ def populate_issues!
     background_populate_issues(self.id)
   end
 
-  def self.exists_with_name?(name)
+  def self.exists_with_name?(name, repos)
     user_name, repo_name = name.downcase.split(?/)
-    Repo.exists?(user_name: user_name, name: repo_name)
+    repos.map { |r| r.full_name }.include?("#{user_name}/#{repo_name}")
```
prathamesh-sonpatki: The original query was also a single query ->

``` ruby
>> Repo.exists?(user_name: 'c', name: 'd').to_sql
  Repo Exists (0.6ms)  SELECT  1 AS one FROM "repos" WHERE "repos"."user_name" = $1 AND "repos"."name" = $2 LIMIT 1  [["user_name", "c"], ["name", "d"]]
```

Here, we are doing calculation in Ruby but I think after adding indexes, the query performance will be better. So I guess original code will perform better with indexes.

app/views/repos/new.html.slim
chrisccerami: Well, the reason I'm passing the repos is so we only need to access the db once. If we call Repo.all each time the method is called, it's eliminating any benefit and it would probably better to just go back to the original implementation. Those are just my thoughts.

app/models/repo.rb
chrisccerami: It's doing one query for each of the user's repos, as opposed to one query total.

Issue #335, opened 2015-05-04
Exception when submitting a duplicate Repo, rather than error message


chrisccerami:
Similar to #314, page crashes when user tries to submit a repo that's already been submitted rather than re-rendering the page with an error message.

chrisccerami:
I've got a few failing tests in this PR that I'm still trying to correct.

chrisccerami:
Tests are passing now. I was having an issue with the test that expect the mailer to not send mail for a repo without issues. I had to rerecord the cassette for that request, and I guess the repo that was previously being used now has issues on it, so I switched it to use a different repo and recorder that response to the cassette. Everything should be good now.

prathamesh-sonpatki:
Closing this since #336 got merged.

chrisccerami:
Thanks. Totally forgot about this issue.

PR #336, opened 2015-05-04
Fixed page crash on duplicate repo

chrisccerami:
Previously when submitting a repo that had already been added, the page would crash. It now correctly renders the page with error messages. #335


schneems:
The error we're seeing:

```
 undefined method `errors' for nil:NilClass
```

This looks like `@repo_sub` is unexpectedly nil. This is happening because the `new` method does not assign a `@repo_sub` variable. To fix we can add `@repo_sub = RepoSubscription.new` in `def new`

chrisccerami:
I have fixed that error in my local version, but I'm dealing with another error so I haven't pushed the change yet. I'll the post the error in a bit.

schneems:
Looks like tests are passing now. You good to go here or you wanted me to wait?

chrisccerami:
@schneems Yeah, should be good to merge. I created this PR as a branch off of 333, so once it gets merged I think that PR should be good to go as well.

CjayBillones:
@chrisccerami does your error messages look just like this?

plain-error-message

This is what I fetched from you branch and if ever you haven't made any designs for it yet, I made it into something like this.

designed-error-message

chrisccerami:
Those were the errors that were already there, I simply fixed the exception that was breaking the page.

CjayBillones:
@chrisccerami Yes yes, I just wanted to confirm if you've done something on the UI of the error message (I might have missed it). :) Thanks!

@schneems @prathamesh-sonpatki I would like to know the status of this PR. I've fetched the fix applied by chris made some changes on the UI and fixed some other bugs that I have found once chris's PR is applied. Will it be fine if I submit my own PR including chris's fix on the exception for you guys to check?

Issue #339, opened 2015-05-12
Can't delete user account via 'Delete Account' link from 'Code Triage misses you' email


ahmetabdi:
http://www.codetriage.com/users/unsubscribe/-random-letter-number-combo-here-

"We're sorry, but something went wrong."

schneems:
Did you change your github username? I don't see you in our database

```
irb(main):007:0> User.where(github: "ahmetabdi").first
=> nil
irb(main):010:0> User.where("github ilike '%ahmetabdi%'").first
=> nil
```

ahmetabdi:
@schneems always been the same

chrisccerami:
I haven't looked at the code for this yet, but I suspect it's trying to render a page with data that no longer exists, since you said it looks like the account was deleted. I'll check it out and see if I can find the offending code.

schneems:
> maybe it did delete

Probably. Have you been getting any code triage emails?

chrisccerami:
I'm thinking what happened is @ahmetabdi tried to delete their account twice. If they tried to visit the unsubscribe url after already deleting the account, you'd get an error, because the account isn't found to be deleted. I can submit a pull request to change this so it will render an error explaining that the account wasn't found and may have already been deleted if you'd like.

schneems:
That would be great. If someone tries to delete a user account that doesn't exist, we should either say so, or simply return "it worked!". 

Maybe we could also link to the github issues and ask people to open a ticket if they cannot stop getting emails. Also calling out a changed github username as a likely source of error would be good. There may be a better way to make our deletion code stronger. I don't know if github provides a UUID for users, but if they do we should probably switch to recording and using that for deletion.

schneems:
Should be fixed by #342

PR #342, opened 2015-05-18
Flash error when user repeat unsubscribes

chrisccerami:
If a user clicks an unsubscribe link twice by mistake, instead of seeing a crash page they will be redirected and receive an flash notice. This will let them know that their account wasn't found and that either it was already deleted, or perhaps they changed their GitHub username since signing up. Discussed in #339.


schneems:
Awesome, thanks! closes #339

Issue #347, opened 2015-06-11
Edit Repository Authorization


CjayBillones:
![bug - edit not found](https://cloud.githubusercontent.com/assets/8434127/8098778/bbbd5802-1026-11e5-8fc2-1bcad8332d90.png)

What happens is that I can't edit the repository even if I own it.

CjayBillones:
I already found out what's wrong with this and I can apply the necessary solution to fix this minor bug. :)

prathamesh-sonpatki:
@CjayBillones Closing this issue as you fixed it already :smile:

CjayBillones:
Alright! Thanks a lot! :)

PR #348, opened 2015-06-11
Fix validation for edit repo authorization

CjayBillones:
Changed the validation checking in able_to_edit_repo? in user model.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -60,7 +60,7 @@ class User < ActiveRecord::Base
   end
 
   def able_to_edit_repo?(repo)
-    repo.user_name == github
+    repo.user_name == github.downcase
   end
 
   def public?
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -346,3 +346,6 @@ DEPENDENCIES
   webmock
   wicked
   will_paginate
+
+BUNDLED WITH
+   1.10.3


CjayBillones:
fixed bug

CjayBillones:
Fix for #347

schneems:
Seems good, thanks

app/models/user.rb
```diff
@@ -60,7 +60,7 @@ def enqueue_inactive_email
   end
 
   def able_to_edit_repo?(repo)
-    repo.user_name == github
+    repo.user_name == github.downcase
```
prathamesh-sonpatki: should both values be downcased?

prathamesh-sonpatki: Ah. My bad. Didn't read issue properly :smile:

CjayBillones: When I checked the code and what's being stored in the database, I observed that the repo.user_name is extracted from github and I think it's already converted to downcase. You might want to double check it to confirm.

CjayBillones: It's okay. Thanks!

And by the way, this line of code from the Repo model might be the ones responsible in making the username and name downcased. :)

screenshot from 2015-06-13 02 37 43

prathamesh-sonpatki: :heart:

PR #349, opened 2015-06-12
Better markdown sanitation

jacarandang:


schneems:
thanks!

schneems:
Ultimately I would like to figure out how to allow links but disallow scripts

jacarandang:
[link name](http://example.com) works for me

Issue #350, opened 2015-06-16
Remove Repository


CjayBillones:
![utility - remove](https://cloud.githubusercontent.com/assets/8434127/8177300/ef8df110-1435-11e5-91c3-35232f94bbee.png)

I think that there still should be a functionality that lets the users remove the repositories that they have added in code triage. In my case, when I first tried the site, I added a 'dummy' repository which doesn't do anything so I don't have any way to actually remove it.

I'll be able to do this functionality. :)

CjayBillones:
HI @prathamesh-sonpatki and @schneems . Have you decided on the implementation of this already? I would like to work on this while I'm on a break from school. :)

schneems:
We haven't started, would love a PR. Let me know if you have any questions. I'm at my parents now any traveling soon but I'll try to be as responsive as possible.

---
Richard Schneeman
http://www.schneems.com




On Sat, Dec 26, 2015 at 1:10 PM -0800, "Cjay Billones" <notifications@github.com> wrote:












HI @prathamesh-sonpatki and @schneems . Have you decided on the implementation of this already? I would like to work on this while I'm on a break from school. :)

CjayBillones:
@schneems just would like to clarify if who do you think should have the authority to remove the repository. The owner of the repository or the one who added the repository?

schneems:
I think anyone who has push access to the repo should be able to delete. 

---
Richard Schneeman
http://www.schneems.com




On Sun, Dec 27, 2015 at 5:03 AM -0800, "Cjay Billones" <notifications@github.com> wrote:












@schneems just would like to clarify if who do you think should have the authority to remove the repository. The owner of the repository or the one who added the repository?

mvz:
It would be really nice if in addition to this users can suggest repos for removal. There are now several repos showing up that are outdated or where no development will take place and the owner is no longer active. The chances of any triaging being useful there are slim, but also the repo owner will probably not actively remove their repo from codetriage.

Examples:
* http://www.codetriage.com/ryanb/cancan (superseded by Cancancan)
* http://www.codetriage.com/opscode/chef (github repo link redirects to chef/chef, which is separately present on codetriage)

boykoc:
I agree with @mvz there may be some cases where there is an outdated repo and triagers could suggest it to be removed to help keep the list of repos up to date as the owner may not be aware/remember to remove it.  
An example includes:

- https://www.codetriage.com/seanlf/jazzify which has been replaced by an up to date version.

PR #351, opened 2015-06-16
Add remove repository

CjayBillones:
Allows users to have the option of removing their repositories from codetriage.

#350 

It looks like this:

screenshot from 2015-06-16 21 14 08


CjayBillones:
I've realised that I've committed some mistakes in this pull request so I'm closing it. I'll be sending a new pull request after I've fixed those mistakes.

PR #353, opened 2015-06-17
Add remove repository

CjayBillones:
utility - remove

This is just the same as #351 but I fixed some backend issues. I just also fixed the cascading delete when removing a repository from code triage.

### DIFF
--- a/test/unit/email_decider_test.rb
+++ b/test/unit/email_decider_test.rb
@@ -2,8 +2,9 @@ require 'test_helper'
 
 class EmailDeciderTest < ActiveSupport::TestCase
   test "decides email frequency" do
-    seed_array = [1,2,3,4,5,6,7]
-    # daily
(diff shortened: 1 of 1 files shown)


prathamesh-sonpatki:
@CjayBillones I think only the person who added the repo should be able to delete it.

Also I am not sure if we should delete repo if there are already some subscribers to it without notifying them. Would love to hear thoughts from @schneems

CjayBillones:
@prathamesh-sonpatki Oh I see. Now that you have mentioned it, I haven't considered the 'authority' of who will be able to remove repos. I would also love to hear your thoughts about this so that I could improve my work. :)

CjayBillones:
I made a new PR, #369 , since I separated this feature into another branch. Sorry for this!

@prathamesh-sonpatki @schneems

CjayBillones:
I'm thinking that the ones that has the authority to remove a repository is the owner himself regardless of whom added it in codetriage. Then subscribers will be notified through e-mail about the deletion of the repository.

PR #355, opened 2015-06-18
Fixed page crash (Merged with chrisccerami's) and other resulting resulting bugs

CjayBillones:
I merged @chrisccerami 's fix from #336 and it is working just fine. I just modified the UI of the error message.

Made it look like: 
designed-error-message

From:
plain-error-message

Also, after fixing the crash, I found this bug wherein after rendering the same page, the starred and the watched repos are cannot be seen.

This was the bug:
bug-starred and watched repos missing

And here's what it looks like after the fix:
fixed bug - starred and watched repo missing

Would like you to check this out @schneems and @prathamesh-sonpatki.


CjayBillones:
I'm trying to figure out why the Travis CI build failed. I tried running the rake test in my local pc and there were no problems.

screen shot 2015-06-22 at 12 39 07 pm

Issue #306, opened 2015-01-15
Add Search for repo on CodeTriage.


AnneTheAgile:
Does Rails provide any built in search tool ? Or to search for a repo here  could we use the Github api to search? 

If I knew how to implement I would look at giving a PR. ty!
AnneTheAgile
p.s. eg. was looking for https://github.com/opscode/knife-azure

prathamesh-sonpatki:
@AnneTheAgile We can start by using pg-search gem I think.https://github.com/Casecommons/pg_search

marceloboeira:
@prathamesh-sonpatki  wouldn't be nice to have a "front-end only" feature to do this, JS based. Using JSON callbacks from github API to search for repos.

prathamesh-sonpatki:
@marceloboeira That would be nice too :smile:

amitguptagwl:
Any update on search feature?

PR #356, opened 2015-06-22
Added search form for repos. #306

GabrielSandoval:
image

I added a search form for the repositories. This form will query through both the repository name and description of the repo table. Query results will be ordered by name, then by the number of issues in descending order, then by description.


bishisht:
I think we should get this going on faster

nateberkopec:
Stale, feel free to reopen.

Issue #362, opened 2015-06-24
Some email's delivered field not being set to true


jacarandang:
When the user subscribes to a repo or clicks "Send new issue!", an email will be sent to them with a random issue. However, the delivered field of this issue assignment is not set to true even if the email was delivered. Since this issue assignment is still marked as not delivered, it will be included in the daily triage email.

PR #363, opened 2015-06-24
Mark email sent by subscribing to a repo and Send new issue! as delivered #362

jacarandang:


app/models/repo_subscription.rb
```diff
@@ -20,6 +20,7 @@ def self.for(repo_id)
     IssueAssigner.new(repo_sub.user, [repo_sub]).assign
     if assignment = repo_sub.user.issue_assignments.order(:created_at).eager_load(:repo_subscription)
                             .where(repo_subscriptions: { repo_id: repo_sub.repo_id }).last
+      assignment.update(delivered: true)
```
prathamesh-sonpatki: Let's use `update!` instead so that we know if it fails.

prathamesh-sonpatki:
@jacarandang Can you add a test case for this?

jacarandang:
For reference #362 
Will add tests tomorrow . Thanks

prathamesh-sonpatki:
:heart:

PR #369, opened 2015-06-25
Add remove repository

CjayBillones:
My previous PR #353 was closed because I cleaned up my master in my forked repository to re-align it to the main repository's master.

### DIFF
--- a/app/controllers/repos_controller.rb
+++ b/app/controllers/repos_controller.rb
@@ -51,6 +51,35 @@ class ReposController < RepoBasedController
     end
   end
 
+  def remove
+    @repos = []
+    @repo = Repo.new(user_name: params[:user_name], name: name_from_params(params))
+
+    Repo.all.each do |x|
+      response = GitHubBub.get("/repos/#{Repo.first.user_name}/#{Repo.first.name}", { token: current_user.token })
+      @parsed_response = JSON.parse(response.body)
+
+      if @parsed_response["permissions"]["push"]
+        @repos.push(x)
+      end
+    end
+  end
+
+  def destroy
+    
+    repo = Repo.find(params[:id])
+    repo_subscriptions = RepoSubscription.where(repo_id: repo.id)
+
+    repo_subscriptions.each do |x|
+      user = User.find(x.user_id)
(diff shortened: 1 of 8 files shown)


prathamesh-sonpatki:
@schneems Thoughts on removing repos? I would allow this only to person who added the repo.

schneems:
Sorry for dropping the ball here. We should have some kind of checks in this method. I would say either the person who added the repo or any of the owners of the repo (on github) should be able to delete it from codetriage. The first case is the easiest.

On a side note, I think you can grab username by calling first on the split array, it would be a little cleaner.

```ruby
@repo = Repo.find_by(name: params[:name].downcase, user_name: params[:url].split("/").first.downcase)
```

CjayBillones:
screenshot from 2016-01-05 18 56 40

I added a remove repo link in the navbar. It basically redirects to a page similar to the Add Repo page but it only lists repos that the current user has push access to (as indicated by @schneems  in #350 ).

I also set-up an email notification that informs all subscribed users that the repo has been removed. I would love to hear your thoughts @prathamesh-sonpatki @schneems .

CjayBillones:
Would like to hear your thoughts about my PR @schneems . :D Thank you!

app/controllers/repos_controller.rb
```diff
@@ -51,6 +51,35 @@ def update
     end
   end
 
+  def remove
+    @repos = []
+    @repo = Repo.new(user_name: params[:user_name], name: name_from_params(params))
+
+    Repo.all.each do |x|
```
schneems: This is going to be pretty expensive. Every time someone clicks that button we have to make 1,845 calls to the github API. I don't think we need a custom view to see all repos that you can remove, I would rather we only had a button on each page. We should probably store the push access data somewhere. Maybe a HABTM table that stores users that have push access that are also code-triage members. I don't know the best way to get this info, i'm thinking we should populate when a new user joins and maybe sync once a day when we do the issue sync.

schneems: Ping, how does that sound?

nateberkopec:
Stale, feel free to reopen.

PR #372, opened 2015-06-29
Modify repo_subscriptions table

CjayBillones:
Removed the user_name and repo_name columns since they are not being used and the table already contains user_id and repo_id which references the corresponding user and repo for that particular repo_subscription.


prathamesh-sonpatki:
Thanks @CjayBillones !

PR #377, opened 2015-06-30
Minor bug fix: Condition on empty starred repos on repos/new page.

GabrielSandoval:
OLD repos/new: 
screen shot 2015-06-30 at 10 23 02 am

NEW repos/new:
screen shot 2015-06-30 at 10 23 19 am

Table now shows "None" when starred_repos is empty.

### DIFF
--- a/app/views/shared/_add_repos_list.html.slim
+++ b/app/views/shared/_add_repos_list.html.slim
@@ -1,5 +1,7 @@
-- if repos.blank?
-  p None!
+- if repos.blank? || repos.count == 0
+  table.table.table-striped
+    tr
+      td None
 - else
   table.table.table-striped
     - repos.each do |repo|


app/views/shared/_add_repos_list.html.slim
```diff
@@ -1,5 +1,7 @@
-- if repos.blank?
-  p None!
+- if repos.blank? || repos.count == 0
+  table.table.table-striped
+    tr
```
prathamesh-sonpatki: @GabrielSandoval I am a bit confused. 

```ruby
[].blank? #=> true
```

So do we need to call count?

GabrielSandoval: Hi @prathamesh-sonpatki ,

To be honest, I was a bit confused too. I just figured out that the repos variable in the code above is not an array but an instance of the SortedRepoCollection class. That's why even though repos doesn't have a content, 'repos.blank?' returns false.
For instance:

screen shot 2015-06-30 at 1 18 26 pm

Where @starred_repos will be passed:
screen shot 2015-06-30 at 1 18 55 pm

ReposController
screen shot 2015-06-30 at 1 19 23 pm

prathamesh-sonpatki: I think we should use https://github.com/codetriage/codetriage/blob/master/lib/sorted_repo_collection.rb#L12 instead of count. Otherwise this looks :+1:

prathamesh-sonpatki:
I merged it manually and pushed to master. Thanks @GabrielSandoval !

PR #379, opened 2015-07-29
Refactor User.find_for_github_oauth method

ck3g:
Extract logic into separate class and split it into smaller methods.


app/services/git_hub_authenticator.rb
```diff
@@ -0,0 +1,57 @@
...
+    @current_user = current_user
+  end
+
+  def self.authenticate(auth, current_user = nil)
+    new(auth, current_user).authenticate
+  end
+
+  def authenticate
+    if user = find_user
+      user.update_attributes(github_params)
```
prathamesh-sonpatki: We should use `update_attributes!` here as we are not checking return value of this statement.

app/services/git_hub_authenticator.rb
```diff
@@ -0,0 +1,57 @@
...
+      User.create(user_params)
+    end
+  end
+
+  private
+
+  attr_reader :auth, :current_user
+
+  def find_user
+    @user ||= current_user ||
```
prathamesh-sonpatki: Do we need caching here as `find_user` is getting called only once?

app/services/git_hub_authenticator.rb
```diff
@@ -0,0 +1,57 @@
...
+      User.find_by(github: auth.info.nickname) ||
+      User.find_by(email:  auth.info.email)
```
prathamesh-sonpatki: I think it should be `auth_email` instead of `auth.info.email` :smile:

app/services/git_hub_authenticator.rb
```diff
@@ -0,0 +1,57 @@
...
+    @token ||= auth.credentials.token
+  end
+
+  def github_email
+    return auth_email if auth_email.present?
+    GitHubBub.get("/user/emails", token: token).json_body.first
+  end
+
+  def github_params
+    @github_params ||= {
```
prathamesh-sonpatki: I think here also we can remove caching as this method is called only once.

app/services/git_hub_authenticator.rb
```diff
@@ -0,0 +1,57 @@
...
+      github:              auth.info.nickname,
+      github_access_token: token,
+      avatar_url:          auth.extra.raw_info.avatar_url
+    }
+  end
+
+  def user_params
+    @user_params ||= github_params.merge(
```
prathamesh-sonpatki: Same here. Here also we can remove caching.

prathamesh-sonpatki:
Good work @ck3g Please check the comments :)

app/services/git_hub_authenticator.rb
ck3g: Good catch =) Thanks

app/services/git_hub_authenticator.rb
ck3g: It's also called from `user_params` as well. `user_params` use that as foundation

ck3g:
Done. I've updated last commit

PR #384, opened 2015-08-25
Cleanup users paths

igas:
We had 3 duplicated routes to `sign_up`. Also we had `/users` and `/users/new` without actual logic in controller. It's not trivial to remove unused user creation, thats why `devise_for` looks like this now, but it's still keeps `:registerable` in model.


schneems:
Looks like the tests are failing, can you take a look?

igas:
@schneems Fixed now. Sorry about that =)

schneems:
Thanks :heart:

PR #397, opened 2015-10-09
Small design tweaks based on thoughtbot designers feedback

kylefiedler:


app/views/users/show.html.slim
```diff
@@ -14,21 +14,32 @@
...
+                = link_to "Go private", user_registration_path(user: {private: true}), method: :patch
+              - else
+                | Private, 
+                = link_to "Go public", user_registration_path(user: {private: false}), method: :patch
+
+        li.slats-item
+          .slats-hgroup
+            h5.slats-title Max number of issues you want to receive per day:
+          .slats-action
+            p #{@user.daily_issue_limit.presence || 'not set'}
```
joshuaogle: This works but you don't need interpolation so it might make more sense as:
```
p= @user.daily_issue_limit.presence || 'not set'
```

(there's one below too)

joshuaogle:
Small suggestion, but looks great after tests are fixed :+1:

PR #398, opened 2015-10-12
Redesign Codetriage

kylefiedler:


kylefiedler:
@schneems this is the branch with all the design changes. I broke the filter on the index page. Do you want me to roll all of these commits into one?

PR #407, opened 2015-10-21
Speed!

schneems:
- Use multi_fetch_fragments for fast home page rendering
- Add pagination so we don't render ALL the repos
- Avoid needless partial calling
- Add javascript ajax based language selection, though now that I think about it this probably would have been a perfect use case for turbo links.
- Speed


arthurnn:
SPEEEEEED

schneems:
This helped a ton: https://github.com/codetriage/codetriage/commit/facfbd775513609024465d70ebd77a14042059e7

PR #418, opened 2015-11-22
Remove puts from user model

prathamesh-sonpatki:


prathamesh-sonpatki:
@schneems Found out this while upgrading to 4.2.5. Might be leftover puts statement.

schneems:
That is for debugging, we need to leave it in. There were bugs with the algorithm that chooses when to send an email. Since it takes days/weeks for the problems to show up I want to be able to look back in time via papertrail and see what the bad values were that caused the incorrect send. I __think__ i've fixed all of them, but it's pretty cheap to leave that in there. It should only show up in the worker.

prathamesh-sonpatki:
:+1:

PR #420, opened 2015-11-24
Add style for adding badges

kylefiedler:
screenshot 2015-11-24 16 17 16

@schneems what do you think of this design instead of having it in a modal?


schneems:
I think it looks good. I'm worried about clutter and call to actions. It might be annoying to always see that call to action if you've already got a badge on your the repo's readme. I based the modal off of Travis like https://travis-ci.org/schneems/wicked, though that behavior might be a bit too hidden. We can wait and see. I'm good with moving forwards with this design.

A smaller window squishes things a bit:



Your CSS is a million times cleaner than mine :smile:

kylefiedler:
@schneems Is that a pattern that all developers know about? Is there another example you can show me like the Travis one with a modal?

schneems:
This project has lots of badges https://github.com/codegram/futuroscope

- Rubygems: New page
- Travis: Modal
- Code Climate: New Page
- Coveralls: Tooltip
- Gemnasium: Couldn't find it
- Gitter: Couldn't find it

So it looks like the most popular is putting it on a new page, the modal wasn't used anywhere else unless you count the tooltip.

PR #421, opened 2015-11-25
Update 404 to point to the correct repo page.

blainesch:
Potentially we could just remove the link as well, if that makes more sense.


schneems:
Thanks!

Issue #412, opened 2015-11-03
Can't search for C# projects


frickinjason:
It looks like the # isn't making it and only C projects are being returned.

schneems:
This is an implementation artifact. You can manually search http://www.codetriage.com/?language=C%23

We need to switch the homepage to use something like turbolinks instead of trying to get all fancy with ajax (is what i'm thinking right now).

prathamesh-sonpatki:
Closing this as #425 got merged.

PR #425, opened 2015-12-04
URI encode filter value for query string

RupW:
so that 'C#' is mapped to 'C%23' for language searches (issue #412)

I know in the issue you proposed changing the UI framework but it's simple enough to patch up the existing AJAX too.


schneems:
Thanks!

PR #426, opened 2015-12-13
Checked in schema.rb changes

prathamesh-sonpatki:
- Running `rake db:migrate` locally has changed `schema.rb` again.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,10 +11,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151021160718) do
+ActiveRecord::Schema.define(version: 20151214120117) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -15,7 +15,6 @@
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "pg_stat_statements"
```
prathamesh-sonpatki: @schneems I just realised that this might be added from Heroku dump and we need it for profiling. Should we add a migration to create this it it does not exist? So that running `rake db:migrate` locally will not show changes.

schneems: Yes please. We should enable stat statements for sure

prathamesh-sonpatki: @schneems Done in https://github.com/codetriage/codetriage/commit/87cb81047a9268abffb7030cb849abbb2dc25bed

prathamesh-sonpatki:
I rebased this with master. Merging it.

Issue #422, opened 2015-11-25
Huge faces on mobile


schneems:
When you view the main repo site on a mobile you get a phone full of huge faces 

![](https://www.dropbox.com/s/ky0eo7rcjqeicq7/Screenshot%202015-11-24%2016.58.50.png?dl=1)

cc/ @kylefiedler

radubogdan:
Interesting. Is someone working on this? @schneems

radubogdan:
Anybody here? 

Looks like [this code](https://github.com/codetriage/codetriage/blob/master/app/assets/stylesheets/components/_avatar.scss#L12) doesn't work anymore.

The `calc()` doesn't return before compiling into css and so the media query looks like: `@media screen and (min-width: 25rem) and (max-width: calc(43rem - 1px)) { ... }` which is not valid CSS.

Also opened a possible issue [here](https://github.com/thoughtbot/bourbon/issues/809)

schneems:
Looks like @kylefiedler is a bit busy. Can you take a stab at a fix? Sorry for the delay

radubogdan:
Sure @schneems. Do you think changing rem to equivalent px value is okay?

schneems:
I'm not an expert here. Would that stay responsive?

radubogdan:
Sure. It has nothing to do with responsiveness.

There is an entire flame on web and I found this [good explanation](http://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css) about px vs rem/em. I personally prefer px.

It's also a good move to go back on what @kylefiedler did when [he first created avatar partial](https://github.com/codetriage/codetriage/commit/cfb81693bd966a412a607a7564ab26850b7ae62b) i.e. using something like `@include media(min-width 25rem max-width 42.9375rem) {`.

I can make a PR with a reference to this issue later today so you can see how it looks using px.

kylefiedler:
@radubogdan Thanks for taking this over. I've been bogged down with other stuff. Since everything else is in `rem`s my preference would be to keep using them.

radubogdan:
@kylefiedler I got to respect that :+1:

PR #430, opened 2016-01-08
Change from calc() to explicit value inside media query (Fixes: #422)

radubogdan:
iPhone 6+

screen shot 2016-01-08 at 20 33 04

iPhone 5

screen shot 2016-01-08 at 20 33 13

Sorry for being a narcissist. I really like the picture :trollface:


schneems:
Thanks! Sorry for this taking so long

PR #442, opened 2016-01-28
Order all_languages alphabetically

andrew:
Ordering all languages alphabetically makes finding your favorite languages in the preferences much easier, before:

screen shot 2016-01-28 at 6 38 08 pm

After:

screen shot 2016-01-28 at 6 52 31 pm


schneems:
Seems like a good idea, thanks!

PR #455, opened 2016-03-16
Explicitly disable caching for Pages#index

### LINKED ISSUE #437: JSON shown when reopening closed tab
Steps to reproduce:

 - open with firefox/chrome (tested on Chrome 49.0.2618.8 and FF 43.0.4) with at least 2 tabs
 - go to http://www.codetriage.com/, select a language (e.g. PHP)
 - url becomes http://www.codetriage.com/?language=PHP
 - close tab
 - reopen last tab

now it shows the JSON response from the server instead of the HTML of the page.
I know it's an edge case but reported anyway :wink:, just in case.

yez:
@schneems 
This should resolve #437.

I have added headers to the response of the index action in the
  pages controller which prohibits the browser from caching the page.

A new test in `pages_controller_test.rb` has also been added.

Tested with Chrome on OS X

### DIFF
--- a/test/functional/pages_controller_test.rb
+++ b/test/functional/pages_controller_test.rb
@@ -19,4 +19,11 @@ class PagesControllerTest < ActionController::TestCase
     get :index
     assert_nil assigns(:repos_subs), 'not assigns to repos_subs'
   end
+
+  test "cache control for headers" do
+    get :index
+    assert_equal("no-cache, no-store, max-age=0, must-revalidate", @response.headers["Cache-Control"])
+    assert_equal("no-cache", @response.headers["Pragma"])
+    assert_equal("Fri, 01 Jan 1990 00:00:00 GMT", @response.headers["Expires"])
+  end
 end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,4 +1,6 @@
 class PagesController < ApplicationController
+  before_action :set_cache_headers, only: [:index]
+
   def index
     @repos = Repo.with_some_issues
     if language = valid_params[:language] || current_user.try(:favorite_languages)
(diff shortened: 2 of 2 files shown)


test/functional/pages_controller_test.rb
```diff
@@ -19,4 +19,11 @@ class PagesControllerTest < ActionController::TestCase
     get :index
     assert_nil assigns(:repos_subs), 'not assigns to repos_subs'
   end
+
+  test "cache control for headers" do
+    get :index
+    assert_equal(@response.headers["Cache-Control"], "no-cache, no-store, max-age=0, must-revalidate")
```
prathamesh-sonpatki: Can you flip the arguments? the expected value should come first and then actual.

yez: Sure, taken care of.

schneems:
Looks like a good idea. @prathamesh-sonpatki you happy?

prathamesh-sonpatki:
Looks good. But do we need all headers. As per this http://stackoverflow.com/a/13168815/925734 we only need `no-cache`. @yez Can you check that?

Also I think the headers should be sent only for  HTML request right? I am not sure but we have JSON and HTML formats in index action. 

On the other hand, what do you think about creating separate controller or endpoint for serving JSON? If we do that we don't have to set any headers ( I think :smile: )

schneems:
I'm fine if we send a few more headers than we need. I actually think this would need to be applied for json responses which is what is being cached. I'm fine if there is no browser caching on the homepage i don't think the perf impact will be that great.

We could also switch the pagination and index page to use turbolinks and get rid of our custom ajax stuff as another option.

prathamesh-sonpatki:
> I actually think this would need to be applied for json responses which is what is being cached

Yup, :sweat_smile:

schneems:
I'm going to merge in, let's clean up later if we need to.

schneems:
Thanks for the help :heart: :rocket:

PR #457, opened 2016-03-25
DOCS DOCTOR INTEGRATION!!!

schneems:
This is a 90% finished implementation of porting docs doctor to run inside of codetriage.

I'm going to be out next week, so I didn't want to do anything crazy like try to merge this in on a Friday. I've not tested the email sending, we also probably want to merge the emails to only send one single email instead of two.

This was easier than I expected, but still time consuming.

Btw if you didn't notice we are now running with SSL 🚀

Have a good weekend!


prathamesh-sonpatki:
💯

schneems:
I think the docs are all working on codetriage right now. I'm going to not be too vocal about it for a few days to see if any major bug reports come in. I don't have plans on migrating people's accounts. It's about 500 people. I'm going to keep both services up for awhile. I'm thinking maybe I can email all people on docsdoctor and ask them to migrate themselves over if they still find the service useful.

In the future I think we need a combined email for issues + docs.

PR #459, opened 2016-04-10
Reset to First Page on Language Switch

### LINKED ISSUE #429: paginate then filter breaks project list on home page
To reproduce:

http://www.codetriage.com/

click "30" to go to the last page of the list of projects

choose Coffeescript from "Filter list by Language:" pulldown

Expected result: it shows me some projects in coffeescript

Actual result: it hides all the projects and the pagination buttons

This makes me think that you are not indexing/detecting any coffeescript project, but I look closer, and notice that the url has page=30 in it: http://www.codetriage.com/?page=30&language=CoffeeScript

Suggested solution: go back to first page when filter is changed.

yez:
This Resolves #429

When paginating and then changing languages, the `page` URL parameter persisted.

It seems that it would be better to reset to the first page when moving between
  languages to prevent the following behaviour:

1. Language "A" has > 10 pages
2. Language "B" has < 10 pages
3. User on Page 12 of language "A" selects language "B"
4. User is shown empty page of language "B" results since language "B" results
   can not fill 12 pages

This code simply extends the existing javascript solution.

### DIFF
--- a/app/views/pages/index.html.slim
+++ b/app/views/pages/index.html.slim
@@ -50,6 +50,7 @@ javascript:
         var language = this.getAttribute("data-language");
         $(".types-filter-button").html(language);
         updateQueryStringParam("language", language);
+        removeQueryStringParam("page");
 
         $.ajax({
           url: document.URL,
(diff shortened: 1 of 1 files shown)


schneems:
Seems like a good idea, thanks!

PR #465, opened 2016-04-27
Bump to Rails 5.0.0.beta4

prathamesh-sonpatki:


prathamesh-sonpatki:
cc @schneems

prathamesh-sonpatki:
Merging this

config/spring.rb
```diff
@@ -0,0 +1,6 @@
+%w(
```
schneems: Can we remove spring from our gemfile instead? 🔥 🔥 🔥

schneems:
Thanks!

config/spring.rb
prathamesh-sonpatki: :smile: Sure I will do that

Issue #434, opened 2016-01-12
random 404 not found when trying to triage


ssoulless:
Hi, I have noticed that for some of the projects when you click the "I want to triage" button

For example one of the projects that presents this problem is cancancan
https://t.co/CExehB3w8e

and some others

schneems:
I had this happen once, going back and clicking the button again fixed it and I was able to subscribe. I don't know what caused the problem.

tejasbubane:
Stumbled across this thing today. The issue is that `I want to triage` button still has `/repo_subscriptions?repo_id=1209` link even though I am logged out. (Happens only in normal window where I was logged in at some point of time before, not in incognito).

prathamesh-sonpatki:
Yeah, this issue happens when we are not logged in but click on `I want to triage`. Devise logs you in and then tries to redirect a POST request resulting into 404.

schneems:
We should probably check if you're logged in or not, if you're not keep the button but have it go to signup instead of posting to the subscribe endpoint.

prathamesh-sonpatki:
:+1: @tejasbubane want to take a stab at implementing it?

mbie:
I think that this issue can be closed. I'm still getting codetriage notifications for it :)

/cc @prathamesh-sonpatki

PR #466, opened 2016-05-01
Fix 404 error when anonymous user tries to triage a repo

mbie:
As @schneems suggested in #434 I've changed "I want to triage..." button behavior. 

When user is not signed the button redirects to authorization page and then user is redirected to the repository page. The user must click the button once again, but there is no 404 error page.

Behavior for logged user remains the same.

### DIFF
--- a/app/helpers/repos_helper.rb
+++ b/app/helpers/repos_helper.rb
@@ -1,6 +1,6 @@
 module ReposHelper
   def wanna_triage_button(repo, html_class: "")
-    path = user_signed_in? ? repo_subscriptions_path(repo_id: repo.id) : user_omniauth_authorize_path(:github)
+    path = user_signed_in? ? repo_subscriptions_path(repo_id: repo.id) : user_omniauth_authorize_path(:github, origin: request.fullpath)
     button_to "I Want to Triage #{repo.path}", path, class: "button #{html_class}"
   end
 end
--- a/app/helpers/repos_helper.rb
+++ b/app/helpers/repos_helper.rb
@@ -1,0 +1,6 @@
+module ReposHelper
+  def wanna_triage_button(repo, html_class: "")
+    path = user_signed_in? ? repo_subscriptions_path(repo_id: repo.id) : user_omniauth_authorize_path(:github)
+    button_to "I Want to Triage #{repo.path}", path, class: "button #{html_class}"
+  end
+end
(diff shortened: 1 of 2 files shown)


mbie:
@schneems @prathamesh-sonpatki any thoughts about this PR?

prathamesh-sonpatki:
We can also hide the button if user is not signed in.

schneems:
I think i originally wanted the button to be there and when you clicked it for it to have you create an account and then drop you off at the same page. I don't know why, but that action is really finicky with devise. I've had to fix it multiple times when it randomly broke. Each time I ask "how did this ever work". I think we even have an integration test for the behavior, or at least we used to.

prathamesh-sonpatki:
Yeah, It is a bit tricky with POST request. I think this PR solves in the way you originally wanted. It will sign up / log in user and then redirect back to home page.

app/helpers/repos_helper.rb
```diff
@@ -0,0 +1,6 @@
+module ReposHelper
+  def wanna_triage_button(repo, html_class: "")
+    path = user_signed_in? ? repo_subscriptions_path(repo_id: repo.id) : user_omniauth_authorize_path(:github)
+    button_to "I Want to Triage #{repo.path}", path, class: "button #{html_class}"
```
prathamesh-sonpatki: Can you also add current request's fullpath as `:origin` param to the `user_omniauth_authorize_path` helper - 

``` ruby
origin: request.fullpath
```

Other than looks good.

prathamesh-sonpatki:
Yeah, It is a bit tricky with POST request. I think this PR solves in the way you originally wanted. It will sign up / log in user and then redirect back to home page.

app/helpers/repos_helper.rb
```diff
@@ -0,0 +1,6 @@
...
+  end
+end
```
schneems: I think add the changes mentioned by @prathamesh-sonpatki and add a newline at the end of the file, and we'll be good.

app/helpers/repos_helper.rb
mbie: Fulpath added.

app/helpers/repos_helper.rb
mbie: Fixed.

prathamesh-sonpatki:
Thanks a lot @mbie :smile: Sorry for late reply!

mbie:
@prathamesh-sonpatki no problem. Thanks for merging :)

PR #469, opened 2016-05-16
Repo#find_by_full_name triggers memory bloat w/ :issues include

itsderek23:
It appears that `Repo#find_by_full_name` triggers some memory bloat when called by repos with many issues:

reposcontroller_show

I believe the offender is the `includes(:issues)`, which thankfully, [I don't believe is used by any of the callers](https://github.com/codetriage/codetriage/search?utf8=%E2%9C%93&q=find_repo&type=Code).

Here's a massive one-line PR to address this.

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -170,7 +170,7 @@ class Repo < ActiveRecord::Base
   end
 
   def self.find_by_full_name(full_name)
-    Repo.includes(:issues).find_by!(full_name: full_name)
+    Repo.find_by!(full_name: full_name)
   end
 
   private


prathamesh-sonpatki:
Nice catch!

schneems:
You found it! Great job.

Did you see the commit that introduced that line https://github.com/itsderek23/codetriage/commit/9543d901e394fe9dc30d7061b0e265e7df6057a0 😉

itsderek23:
image

CodeTriage: our codez are so good, we make our own bugs.

schneems:
The bug was in production for a good 6 months-ish, which is how long it took me to track down. I re-introduced it to play with some other debugging tools. None have found them as quickly or accurately as Scout.

PR #472, opened 2016-06-06
Added email frequency to users

GustavoCaso:
#458 #244 

@schneems Added `email_frequency` so the user can select there best preference for receiving email from CodeTriage

screen shot 2016-06-06 at 09 16 09

screen shot 2016-06-06 at 09 17 54

PR #475, opened 2016-06-07
Refine user selectable email frequency

schneems:
We don't want a user setting a "default" value to short circuit our deciding logic. We still want to back off if they're not interacting so they don't get burned out and turn off  emails. 

This PR includes commit from #472


schneems:
Deploying now, thanks!

Issue #473, opened 2016-06-06
Need to truncate titles in DB


schneems:
Updating an issue with a really long title:

```
  SQL (1.7ms)  INSERT INTO "issues" ("url", "last_touched_at", "number", "created_at", "updated_at", "repo_id", "title", "html_url", "state") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING "id"  [["url", "https://api.github.com/repos/fsharp/FAKE/issues/1221"], ["last_touched_at", 2016-04-12 18:34:04 UTC], ["number", 1221], ["created_at", 2016-06-06 22:27:56 UTC], ["updated_at", 2016-06-06 22:27:56 UTC], ["repo_id", 244], ["title", "MSBUILD : error MSB1021: Cannot create an instance of the logger. Could not load file or assembly 'Microsoft.Web.XmlTransform, Version=2.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified."], ["html_url", "https://github.com/fsharp/FAKE/issues/1221"], ["state", "open"]]
```

Gives an error

```
ActiveRecord::ValueTooLong: PG::StringDataRightTruncation: ERROR:  value too long for type character varying(255)
: INSERT INTO "issues" ("url", "last_touched_at", "number", "created_at", "updated_at", "repo_id", "title", "html_url", "state") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING "id"
```

Easiest thing i think is to truncate at 254 chars and ad an ellipsis

prathamesh-sonpatki:
@schneems We can also make the field text. Postgres does not have performance impact. 

Check Tip on this page - https://www.postgresql.org/docs/9.1/static/datatype-character.html

> There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

prathamesh-sonpatki:
For UI, we can use `truncate` or something.

schneems:
I'm fine with either method.

schneems:
I just installed rollbar recently and this is happening quite a bit

```
[...]

PR #478, opened 2016-06-13
Change issues title to text

GustavoCaso:
#473

@schneems I thought the simplest way is to change the field to `text`, because like @prathamesh-sonpatki there is no performance issue.


schneems:
Sorry I missed this when it came in. Thanks!

Issue #494, opened 2016-07-18
Email frequency select box background


gizotti:
Currently the Email Frequency select box has a white background, what makes it look like it has no text select.
It also looks very different in Firefox than it does in Safari and Chrome. I couldn't check IE, but if anyone can do that would be awesome.

Here's how they look at the moment:

Safari:
![image](https://cloud.githubusercontent.com/assets/1103597/16913321/d86e6150-4d2c-11e6-9ace-542a6f748d68.png) ![image](https://cloud.githubusercontent.com/assets/1103597/16913272/9012a2fe-4d2c-11e6-84d1-b8e15212fbab.png)

Chrome:
![image](https://cloud.githubusercontent.com/assets/1103597/16913285/a2262d6c-4d2c-11e6-9214-b3869649da91.png) ![image](https://cloud.githubusercontent.com/assets/1103597/16913289/a836650a-4d2c-11e6-9864-095f451a2bbe.png)

Firefox:
![image](https://cloud.githubusercontent.com/assets/1103597/16913297/b7c6ddba-4d2c-11e6-96a3-c89caac878f0.png) ![image](https://cloud.githubusercontent.com/assets/1103597/16913303/bf03acac-4d2c-11e6-9800-4c1724bc0571.png)

It's a pretty simple one. If you confirm what the layout should be I can get that done.

schneems:
Guess we should default to "daily". Text should be black. A PR would be wonderful

gizotti:
How about the styling of the firefox select? Should I change I make it smaller and make it looks similar to the others?

Issue #432, opened 2016-01-10
Unselected values difficult to read in "Triage More" selector


cbzehner:
Number of items to triage dropdown has no contrast between unselected options value and background. Tested on Linux Mint (Ubuntu-derivative) and Windows 10 on Chrome for repro.

Steps:
1) Login with Github
2) Go to repro and select "I want to triage"
3) Choose selector under "Triage more or less" and see values which display as hard to read.

![codetriage colors](https://cloud.githubusercontent.com/assets/3886290/12223970/78487d42-b79a-11e5-92c2-13338d6f060e.JPG)

schneems:
I can't see that on my mac :( Can you attempt a fix? I think maybe we should explicitly set the color of the text.

cbzehner:
Definitely. I'll look into it this weekend.

mjhoy:
I see it on FF 44.02, osx 10.9

adding the rule

    .repo-instructions select { color: black; }

fixed it for me.

prathamesh-sonpatki:
Fixed by https://github.com/codetriage/codetriage/pull/495

PR #495, opened 2016-07-19
Ensure select text colour is black

gizotti:
This PR adds a css styling across all select boxes to ensure black text colour.

With this PR issue #494 can be closed, and I believe #432 as well.

Select value already defaults to "daily", as mentioned on #494, so nothing else had to be done.


schneems:
Thanks, no clue why builds are failing, but i'm pretty sure it's not due to this PR.

gizotti:
No worries, I made sure I ran it in my local env and all passed, so I guess it was just a hiccup hey.

Issue #482, opened 2016-06-22
Triage Docs Button Fails


scottwhudson:
The triage docs button seems to fail if the user doesn't first subscribe to issues for the same repo.  Just curious, is this expected behavior?

schneems:
it isn't

scottwhudson:
Good to know.  I found the problem here, the triage docs button sets the `:email_limit` default to 0 if the user isn't already subscribed to the repo, which doesn't pass the repo_subscription model validation.  

On the surface, changing the validation looks pretty straightforward, but I haven't really dug into the mailers to see what consequences that might have (seems like validations usually exist for a reason, right?).  It looks like issues and docs behave a little differently, would you recommend going the validation change route or looking for another way to solve this?

schneems:
I think we need to update to allow 0 in the `email_limit`.

scottwhudson:
No problemo!

PR #496, opened 2016-07-20
Enable separate doc registration #482

scottwhudson:
#482 

Fixes bug where users must select issues to triage before selecting docs, and incorporates minor view tweaks to standardize the two.


schneems:
Thanks!

Issue #485, opened 2016-06-23
`foreman start` not working for local setup


anilmaurya:
I cloned codetriage repo, and followed setup guide in Readme, however when I run `foreman start` I see strange behaviour. All process exit with code 0.

Log trace:
```
codetriage (master)*$ foreman start
19:36:57 web.1     | started with pid 16419
19:36:57 worker.1  | started with pid 16420
19:36:57 release.1 | started with pid 16421
19:36:58 web.1     | [16419] Puma starting in cluster mode...
19:36:58 web.1     | [16419] * Version 3.4.0 (ruby 2.3.1-p112), codename: Owl Bowl Brawl
19:36:58 web.1     | [16419] * Min threads: 5, max threads: 5
19:36:58 web.1     | [16419] * Environment: development
19:36:58 web.1     | [16419] * Process workers: 2
19:36:58 web.1     | [16419] * Preloading application
19:37:08 web.1     | I, [2016-06-23T19:37:08.250196 #16419]  INFO -- : Loaded Configuration: /.../codetriage/config/scout_apm.yml. Using environment: development
19:37:08 worker.1  | I, [2016-06-23T19:37:08.251766 #16420]  INFO -- : Loaded Configuration: /.../codetriage/config/scout_apm.yml. Using environment: development
19:37:08 release.1 | I, [2016-06-23T19:37:08.250200 #16421]  INFO -- : Loaded Configuration: /.../codetriage/config/scout_apm.yml. Using environment: development
19:37:09 web.1     | [16419] * Listening on tcp://0.0.0.0:3000
19:37:09 web.1     | [16419] Use Ctrl-C to stop
19:37:09 web.1     | [16419] - Worker 1 (pid: 16430) booted, phase: 0
19:37:09 web.1     | [16419] - Worker 0 (pid: 16429) booted, phase: 0
19:37:10 release.1 | exited with code 0
19:37:10 system    | sending SIGTERM to all processes
19:37:10 web.1     | [16419] - Gracefully shutting down workers...
19:37:10 worker.1  | terminated by SIGTERM
19:37:10 web.1     | [16419] === puma shutdown: 2016-06-23 19:37:10 +0530 ===
19:37:10 web.1     | [16419] - Goodbye!
19:37:10 web.1     | exited with code 0
```
When I run each process in separate console, its working fine.

tejasbubane:
This is because `release` (which runs `db:migrate`) is exiting. 

[...]

PR #497, opened 2016-07-21
[ci skip] Use heroku local instead of foreman

tejasbubane:
Change README to use `heroku local`.
Forman start failed for `release rake db:migrate` command.
Use `heroku local` instead. See #485 for more details.


schneems:
Thanks!

PR #499, opened 2016-07-25
RepoBasedController#find_repo: Automatically Update Repository If updated_at Is Over 6 Hours Ago

rye:
The changes introduced will cause `RepoBasedController#find_repo` to check if the repository was updated over 6 hours ago, and queue an UpdateRepoInfoJob if so.  It returns the result of `Repo.find_by_full_name` either way, and only performs the check and job queuing if the result of this find is truthy.

I'm unsure about the performance drawbacks of this, since this check is performed every time a `RepoBasedController` uses the `#find_repo` method to locate a Repository.  I think this could also be split out into a separate method, and then the new method could be more appropriately used.

I'm willing to work on improvements to this as well.

### DIFF
--- a/app/controllers/repo_based_controller.rb
+++ b/app/controllers/repo_based_controller.rb
@@ -1,11 +1,11 @@
 class RepoBasedController < ApplicationController
-  protected
+	protected
 
-    def name_from_params(options)
-      [options[:name], options[:format]].compact.join('.')
-    end
+	def name_from_params(options)
+		[options[:name], options[:format]].compact.join('.')
+	end
 
-    def find_repo(options)
-      Repo.find_by_full_name(options[:full_name])
-    end
+	def find_repo(options)
+		Repo.find_by_full_name(options[:full_name])
+	end
 end
--- a/app/controllers/repo_based_controller.rb
+++ b/app/controllers/repo_based_controller.rb
@@ -1,11 +1,15 @@
 class RepoBasedController < ApplicationController
-	protected
+  protected
 
-	def name_from_params(options)
-		[options[:name], options[:format]].compact.join('.')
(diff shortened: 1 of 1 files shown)


rye:
Another performance drawback that I can think of is: if the queue of UpdateRepoInfoJobs is too big, it is possible to flood the queue with jobs to update a repo if one makes repeated requests and keeps firing that callback.  It might make sense to check if the queue already has an UpdateRepoInfoJob for a repository before adding another job, or to introduce some kind of lock on the repository that prevents it from repeatedly being spammed with requests to update it.

app/controllers/repo_based_controller.rb
```diff
@@ -1,11 +1,15 @@
...
-    end
+  def name_from_params(options)
+    [options[:name], options[:format]].compact.join('.')
+  end
 
-    def find_repo(options)
-      Repo.find_by_full_name(options[:full_name])
+  def find_repo(options)
+    if repo = Repo.find_by_full_name(options[:full_name])
+      repo.update_repo_info! if repo.updated_at < (Time.now - 6.hours)
```
prathamesh-sonpatki: I don't think it should be present in We can add a cron to update the repos after every X hours.

nateberkopec:
I agree, I don't see why this has to be done during a search and not just something done on a regular scheduled cron. Closing this solution, but raising the problem in issue

mario-d-s:
> It certainly seems like it would be better if you could get specific information for all repos in one request, but I don't think GitHub's API supports obtaining that much information in the public `/repositories` endpoint.

Is using GraphQL an option? I don't have experience with it, nor with GitHub's API, but it looks promising for a use case like this.

PR #501, opened 2016-07-30
[ci skip] Small fixes in the README

tejasbubane:
* Using `Heroku Local` instead of `Foreman`
* Change port numbers to `3000` instead of `5000` as specified in the
  `.env` file instructions.


prathamesh-sonpatki:
Thanks!

Issue #498, opened 2016-07-22
Allow filtering issues


kevincox:
For some projects I am often getting open issues that are in-progress, tracking bugs or other non-fresh issues. It would be really nice if the issues you received were issues that are already un-triaged or ready to be fixed.

I see a couple of things that could be useful here.
- Project defaults: Most projects have a label system and it would be nice to be able to set it once and affect all users.
- User overrides. Maybe a user is only interested in a certain topic. It might be useful to allow the user to override the project default. However since this project is aimed at triage it might not be very useful as there is not much to filter on.

Examples:
- is:open is:issue -label:triaged -label:needinfo no:assignee 
- is:open no:assignee
- is:open comments:0
- is:open "created:>2 days ago"

I think this change would really help aim help where it really matters.

schneems:
Here's things I think would be useful:
- To subscribe to specific labels or restrict specific labels
- Search since last comment time

I rarely see issues with truly zero comments, I don't see the use of searchinf for a specific number or a less or greater than. I could maybe be convinced assignee is an interesting search metric.

I've wanted the ability to search or subscribe via labels for a long time. But it's not a strong want, so i've not implemented it. It's a feature I would consider helping out with though if you're interested in taking a stab.

kevincox:
I see the comment count being not as useful but I think it would be awesome if you could copy a search string from github and use that. I don't know how easy that would be. However supporting common queries and reporting other things being unsupported would be good as well. 

Things I would use for sure if they were avaikavkr:
- Filter on assigned. If I'm triaging I probably don't have to worry if someone is assigned. 
- Filter on labels. I think this is an obvious win. 
- Filter on last updated time. A triager doesn't need to spend on time on issues that are having an active discussion. 

I don't have much time ATM but if no one jumps on this I might try my hand in a month or two.

hmasila:
@kevincox I am not sure how far you went with this, but I decided to give it a try. Have a look and let me know what you think

cc @schneems

kevincox:
I didn't do any more work then shown the in PR. However the discussions didn't go anywhere so I left it unfinished. Thanks for looking into this.

PR #507, opened 2016-09-20
Store labels in the db.

kevincox:
The primary key of repo, label, issue will allow fast filtering of
issues matching a label in a repo.  This is the most email used for
sending emails.

This is progress towards #498

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -4,6 +4,7 @@ class Issue < ActiveRecord::Base
 
   validates :state, inclusion: { in: [OPEN, CLOSED] }
   belongs_to :repo
+  has_many :labels, class_name: :IssueLabel, dependent: :delete_all
 
   after_save    :update_repo_issues_count
   after_update  :update_repo_issues_count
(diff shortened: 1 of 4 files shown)


schneems:
Could we use a postgres array for this? The upside would be less to worry about, the downside is that it would be harder to query all issues against a specific tag. what do you think @prathamesh-sonpatki ?

prathamesh-sonpatki:
We can split this into tables, `labels` and `issue_labels`.

`labels` will have name and repo_id.
`issue_labels` will have `label_id, `issue_id` and `repo_id`.

Then we will be able to show all issues for a particular label, query them and do all sorts of things.

kevincox:
@schneems My first thought was postgres arrays but it depends on how we want to query the data. We should looks back at #498 first and decide what queries we want to support.

@prathamesh-sonpatki I don't know what queries that split will get us over the current situation. It will save a couple of bytes (maybe not actually because of the extra table) but I don't see any benefit over that.

nateberkopec:
I think there's probably a use case for searching for issue labels cross-repo, e.g. `language:ruby label:help-wanted`. This PR could support that with a new index. I think PG arrays are a red herring.

@kevincox Can you rebase?

PR #512, opened 2016-11-13
use https protocol instead of git

lalithr95:
Current git source uses git protocol which is insecure, this PR replaces git sources to use https

Review: @schneems


lalithr95:
bump ?

PR #514, opened 2016-11-18
Added meta tags helper to render default or custom meta tags inside the helper

rubyrider:
Added meta tags helper to render default or custom meta tags inside the helper. This fix will resolve issue#513.

Given default meta tags are:
Name: CodeTriage
Description: Help out your favorite open source projects and become a better developer while doing it

To render custom meta tags:
just use `content_for(:meta_tags)` block and define your meta tags. That's it!

This is how it shows to facebook:
screen shot 2016-11-18 at 12 53 50 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -29,4 +29,10 @@ module ApplicationHelper
   def star_svg
     STAR_SVG
   end
+
+  # Render custom or default meta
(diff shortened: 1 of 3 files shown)


schneems:
Thanks! Would like to eventually make a better one for each project page so codetriage.com/puma/puma has a custom call to action for example.

PR #518, opened 2016-12-16
Ruby 2.4 Compat

schneems:
Not sure why but `bundle install` does not update the `Gemfile.lock` to 2.4.


prathamesh-sonpatki:
@schneems :+1:

PR #520, opened 2016-12-17
Bump coffee-rails and rollbar gems

maclover7:
These two gems are preventing CodeTriage from running CI against Rails
master. In the case of coffee-rails, it had a version restriction that
required the gem version being bumped. In the case of rollbar, the
version ahd to be bumped because the previous version has been using the
now removed `alias_method_chain`.


prathamesh-sonpatki:
Awesome!

PR #523, opened 2017-01-13
Add top border radius to repo item headers.

jacortinas:
Hey there 👋

I noticed a small graphical bug while hovering over repo items. Attached is a gif showing the graphical glitch. When hovering over any `.repo-item`, all other `.repo-item`'s are affected.

codetriage-glitch

It seems that while hovering over a `.repo-item`, the corners of the `.repo-item-header` are showing for a split second. A simple solution is to round the top corners of the `.repo-item-header` to match the `.repo-item`.


prathamesh-sonpatki:
Thanks @jacortinas 🥇

PR #526, opened 2017-01-31
Changes to schema.rb after rake db:migrate on clean setup

elliotthilaire:
I am setting up codetriage for development on my machine for the first time.
I have not had the databases or code before so I have a completely fresh environment.

I have followed the steps in the readme verbatim.
These are the changes to schema.rb that occured after running.
```
bin/rake db:create
bin/rake db:migrate
```

I have not investigated into the cause yet as I don't know much of the history.
I thought it would be of interest to you straight up as it seems a few tables, and defaults have been removed.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -73,8 +73,8 @@ ActiveRecord::Schema.define(version: 20160927195642) do
 
   create_table "issue_assignments", force: :cascade do |t|
     t.integer  "issue_id"
-    t.datetime "created_at",                           null: false
-    t.datetime "updated_at",                           null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
     t.integer  "repo_subscription_id"
     t.boolean  "clicked",              default: false
     t.boolean  "delivered",            default: false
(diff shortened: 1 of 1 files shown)


schneems:
Not totally following. What's the issue here? Do we need to change our readme?

Maybe should be `rake db:schema:load` instead of `db:migrate` ?

elliotthilaire:
The issue is that migrations and schema.rb are out of sync.

As a newcomer setting up the project for the first time I expected to run `rake db:migrate` and see no changes to schema.rb.

Instead two create_table definitions are removed from `schema.rb`, opro_auth_grants and opro_client_apps as, shown in the diff.

Put another way; I expect `rake db:migrate` and `rake db:schema:load` to set up the same tables. In the projects current state, they will not.

Sorry I wasn't clear before.
If you don't see this as an issue we can close it without any change.

schneems:
I would say that `rake db:schema:load` is the preferred way to set up a project that is sufficiently old as some migrations face bit-rot over time. Maybe update the docs to that and I can merge that PR.

elliotthilaire:
No problem. I'll close this now and do that.

PR #527, opened 2017-02-03
Remove reference to docsdoctor.org in docs email

elliotthilaire:
Congratulations on the merging of docsdoctor.org into codetriage.com

This removes a small remnant. docsdoctor.org currently goes to a 'There's nothing here, yet.' Heroku page.


prathamesh-sonpatki:
Good catch 🤘

Issue #528, opened 2017-02-03
Email frequency setting is set to 'Daily' unintentionally when changing other preferences


elliotthilaire:
The select field for Email frequency is not pre-filling with the persons current option.

Steps to reproduce:
1. Click EDIT ACCOUNT
2. Change Email Frequency to 'Once a Week'
3. Click 'SAVE'
4. Observe that 'Once a week' is shown
5. Click 'EDIT ACCOUNT' again
6. Observe that form has reset to 'Daily'
7. Make no changes to the form and click 'SAVE'
8. Your email frequency has been reset back 'Daily'

elliotthilaire:
Resolved with https://github.com/codetriage/codetriage/pull/529

PR #529, opened 2017-02-04
Fix edit page not remembering email frequency

elliotthilaire:
Fixes:
https://github.com/codetriage/codetriage/issues/528


schneems:
thanks!

Issue #244, opened 2014-04-04
Send triage email based on time preference


alksl:
It would be nice to be able to set a preference when the triage email is sent out. For now it's problematic if you are in a different timezone.

I looked around in the source but I could't find where the emails are sent out. Some pointers would be helpful.

elliotthilaire:
I receive emails at 5pm which is about the time I make sure my inbox is clear.
I would like to be able to set the time of day I receive the emails so they arrive when I am most motivated (for me, in the morning).

As this is an old issue. I want to ask before starting; Is this worth exploring?

prathamesh-sonpatki:
Good idea :+1:

elliotthilaire:
I've started work on this over at https://github.com/elliotthilaire/codetriage/tree/add-email-time-of-day-preference

The idea being that a person can select a preferred time of day from a dropdown box with 24 values.  [...,5:00am UTC, 6:00am UTC,...] etc.

We have a EmailDecider class that is used like this:
https://github.com/codetriage/codetriage/blob/master/app/models/user.rb#L123:L133

I am thinking of modifying this so that the Heroku schedular can run the email task hourly, and EmailDecider will cause a will skip if the current time of day is before the persons preferred time of day. (EmailDecider should take take of skipping if a person has already received today's email).

PR #530, opened 2017-02-10
Add email time of day preference

elliotthilaire:
Closes https://github.com/codetriage/codetriage/issues/244

This adds an email time of day setting on the users settings page.

screen shot 2017-02-10 at 2 44 41 pm

screen shot 2017-02-10 at 2 42 08 pm
 
screen shot 2017-02-10 at 2 42 37 pm

The Heroku schedular can be set up to run hourly and `send_daily_triage!` will skip sending the triage if it is before the users set preference.

I can set a default email time of day. How would you like to do this?
1. A default set in the code
2. A rake task to update users

This PR includes extra changes in schema.rb. These changes come from running db:migrate on afresh setup. Details here: https://github.com/codetriage/codetriage/pull/526

Happy to take on feedback and make further changes.


elliotthilaire:
Hello @schneems,  @prathamesh-sonpatki 

A further comment on this. I have not tested this on a live server. I looked at setting up my own Heroku instance and running it for a few days before submitting a pull request.

This is what stopped me:
1. I don't know what/when scheduled tasks are set up live, so I would be guessing about the configuration.
2. I'm not sure if I can it relying only on free services/addons, redis, email, etc.
3. newrelic.yml and skylight.yml have hardcoded keys which might interfere.

Do you have a staging setup this can be tested on?
Would you be willing to assist me in setting up my own?

db/schema.rb
```diff
@@ -73,8 +73,8 @@
 
   create_table "issue_assignments", force: :cascade do |t|
     t.integer  "issue_id"
-    t.datetime "created_at",                           null: false
-    t.datetime "updated_at",                           null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
     t.integer  "repo_subscription_id"
```
MonikaMahanthappa: The default database constraints are getting removed. Unless it is intentional, would suggest that you do not make any changes not related to your story in this PR. Getting it back to as it is on master. 

This issue might have caused due to some Postgres version difference.

test/unit/user_test.rb
```diff
@@ -93,6 +93,91 @@ class UserTest < ActiveSupport::TestCase
...
+    now = Time.now
+    Timecop.freeze Time.new(now.year, now.month, now.day, 02, 0, 0) do
+      assert_difference("User.find(#{user.id}).issue_assignments.where(delivered: true).count", 0) do
+        Issue.any_instance.stubs(:valid_for_user?).returns(true)
+        user.send_daily_triage!
+      end
+    end
+  end
+
+  test 'send_daily_triage! can be scheduled to run multiple times a day' do
```
MonikaMahanthappa: I think that this test is not required/redundant. It is already covered in the above tests.

test/unit/user_test.rb
```diff
@@ -93,6 +93,91 @@ class UserTest < ActiveSupport::TestCase
...
+    repo  = repos(:rails_rails)
+
+    issue = repo.issues.new
+    issue.title           = "Foo Bar"
+    issue.url             = "http://schneems.com"
+    issue.last_touched_at = 2.days.ago
+    issue.state           = 'open'
+    issue.html_url        = "http://schneems.com"
+    issue.number          = (repo.issues.last.try(:number) || 0) + 1
+    issue.save
```
MonikaMahanthappa: It is not required to create the issue with all attributes set as this test is not particularly testing any of those values. Eg, It can be `repo.issues.create(title: 'Foo Bar')` or `issue = repo.issues.new(title: 'Foo Bar'); issue.save`

app/helpers/users_helper.rb
```diff
@@ -0,0 +1,5 @@
+module UsersHelper
+  def available_times_of_day
+    (0...24).map { |hour| Time.utc(2000, 1, 1, hour, 00, 0) }
```
MonikaMahanthappa: To make it easier for the user, I suggest that we take the user timezone also into consideration.

test/unit/user_test.rb
elliotthilaire: Thanks @MonikaMahanthappa. I'll take a look to see how it could be simplified. I'm not sure if I will run into issues where some attributes have to be set before they are saved.
I duplicated the existing tests on send_daily_triage! which have this setup.

test/unit/user_test.rb
elliotthilaire: Thanks @MonikaMahanthappa. The tests above you refer too only run the send_daily_triage! method once.

This test was to ensure that if we call send_daily_triage! multiple times after the preferred time we won't send multiple emails. Since this is the responsibility of EmailDecider I agree it is a redundant.

app/helpers/users_helper.rb
elliotthilaire: Thanks @MonikaMahanthappa, I agree.  I wanted to create the simplest PR possible so I left out user time zones. 

If we present the times as UTC and save the times as UTC we have a working feature and a solid base to build user time zones on.

Do you think this approach is ok?

elliotthilaire:
I've updated this with feedback from @MonikaMahanthappa. Thanks Monika.

I've also added a default time of 5pm UTC.
This is so that the scheduled task can run hourly without a change for when people receive their emails.
This default was chosen based on when I have been getting my emails.

nateberkopec:
> I have not tested this on a live server. 

@elliotthilaire can you rebase this? I think review apps are set up now, so you should be able to test this live if you rebase and push.

elliotthilaire:
Hi @nateberkopec,

I've rebased and pushed. 
Heroku CI has failed but I'm unable to see why.

schneems:
If travis is happy then I am too. There's some wonkyness going on with the Heroku suite. I migrated the app to an org and then back again and Heroku did not like that at all. 

If you can address the issues Nate touched on I think we can get this over the line.

elliotthilaire:
I am working through conflicts introduced by the refactor:
https://github.com/codetriage/codetriage/pull/554/files

I am planning on adding display of local timezone using gem 'local_time'

schneems:
Builds are working. What else is left?

> I am planning on adding display of local timezone using gem 'local_time'

Is that it?

nateberkopec:
@schneems I'd like to another code review round before this gets merged if that's OK w/you

config/initializers/new_framework_defaults.rb
```diff
@@ -20,3 +20,6 @@
 
 # Do not halt callback chains when a callback returns false. Next major version defaults to false.
 ActiveSupport.halt_callback_chains_on_return_false = true
+
+# Time columns will become time zone aware in Rails 5.1
+Rails.application.config.active_record.time_zone_aware_types = [:datetime, :time]
```
elliotthilaire: I think this can go now that we are Rails 5.1

elliotthilaire:
Builds are working. What else is left?

       I am planning on adding display of local timezone using gem 'local_time'

   Is that it?

@schneems Yes.

gem 'local_time' does not work with our version of turbolinks.
Pointing gem file to the local_time 2.0 branch in github fixes this.
https://github.com/basecamp/local_time/pull/54

I can't use local_time as a drop in on top of the select box options because the display values can't include html tags. I need to figure out a nice way to do this.

schneems:
Looks like we can use https://github.com/basecamp/local_time/pull/55

Can you try out

```
gem 'local_time', git: 'https://github.com/basecamp/local_time', branch: '2-0'
```

elliotthilaire:
> Looks like we can use basecamp/local_time#55

Yep. That's what I did to solve the first issue.

The second issue is that you cannot do `<option value="..."><%= local_time(time) %></option>`
So I need to find another way.

elliotthilaire:
I think this is done.

The server doesn't know about the users timezone. The times are always sent to and from in UTC, and stored in UTC. If javascript is enabled. It will replaced the display of the times to the local time.

screen shot 2017-06-03 at 4 22 50 am

screen shot 2017-06-03 at 4 23 49 am

The order of the select box does not change. They are ordered by value, not text. I don't think this should be a blocker for this PR. It might be more suitable for someone with better JS fu than I.  I did have a look into reordering but wasn't happy with the solutions I could come up with.
 
screen shot 2017-06-03 at 4 31 35 am

elliotthilaire:
Happy for another round of review but am confident if you merge.
Once merged, you can change Heroku to run the scheduled task each hour.

elliotthilaire:
I've rebased and fixed conflicts. Then squashed to cleanup history.

app/jobs/send_daily_triage_email_job.rb
```diff
@@ -28,4 +28,14 @@ def skip_daily_email?(user)
   def email_decider(user)
     @email_decider ||= EmailDecider.new(user.days_since_last_clicked, minimum_frequency: user.email_frequency)
   end
+
+  DEFAULT_EMAIL_TIME_OF_DAY = Time.utc(2000, 1, 1, 17, 00, 0)
+
+  def before_email_time_of_day?(user)
```
nateberkopec: Comparing times as strings seems weird. Why not just compare the times directly?

test/jobs/send_daily_triage_email_job_test.rb
```diff
@@ -8,20 +8,64 @@ def setup
...
   test 'does not deliver if no subscriptions' do
     def @user.issue_assignments_to_deliver; []; end
+    def @job.before_email_time_of_day?(*); false; end
 
     refute @job.perform(@user)
   end
 
   test 'does not deliver if email should be skipped' do
     def @user.issue_assignments_to_deliver; IssueAssignment.all.limit(1); end
+    def @job.before_email_time_of_day?(*); false; end
```
nateberkopec: no need for this line because skip_daily_email? returning true will short circuit the conditional

test/jobs/send_daily_triage_email_job_test.rb
```diff
@@ -8,20 +8,64 @@ def setup
...
     def @job.skip_daily_email?(*); true; end
 
     refute @job.perform(@user)
   end
+
+  test 'when email_time_of_day not set, it delivers after default time of day' do
+    def @user.issue_assignments_to_deliver; IssueAssignment.all.limit(1); end
+
+    Timecop.freeze time_preference_for_today(SendDailyTriageEmailJob::DEFAULT_EMAIL_TIME_OF_DAY) + 1.hour do
```
nateberkopec: Rather than use a new testing library, we can just use [Minitest stubs](https://github.com/seattlerb/minitest/blob/master/lib/minitest/mock.rb#L212).

Time.stub(:now, your_time) do ...

nateberkopec:
Thanks for seeing this through!

app/jobs/send_daily_triage_email_job.rb
elliotthilaire: The time column stored the time as well as the data 2000-1-1. 
```
irb(main):002:0> User.first.email_time_of_day.utc
 => 2000-01-01 17:00:00 UTC
```
This is comparing the time it is now for today, against the time that is set in the database (ignoring the data part).

If you know of a method to handle this, I'm open for suggestions.

It's why you also see this further down
```
Time.utc(now.year, now.month, now.day, time.hour, time.min, time.sec)
```

test/jobs/send_daily_triage_email_job_test.rb
elliotthilaire: That's true. 

If I remove this line and then and `skip_daily_email?` develops a bug.
Then this will pass with a false positive depending on the time of day it is run.

Correct me if I'm wrong.

elliotthilaire:
> Thanks for seeing this through!

@nateberkopec, No problem. 

I've left comments for 2/3 of the requested changes.

test/jobs/send_daily_triage_email_job_test.rb
elliotthilaire: I am wrong, because `skip_daily_email?` is mocked.

elliotthilaire:
I've rebased and changed my refutes to assert_nots based on. https://github.com/codetriage/codetriage/commit/789a6a74206728486ea7d05172c772976b594443

nateberkopec:
> I've rebased and changed my refutes to assert_nots based on

Thank you! @schneems told me to do that.

app/jobs/send_daily_triage_email_job.rb
nateberkopec: Let's just convert time.current:

```ruby
Time.current.change(:month => 1, :day => 1, :year => 2000)
```

nateberkopec:
One last comment about code, but otherwise LGTM. Let me figure out how to get this on a review app...

app/jobs/send_daily_triage_email_job.rb
elliotthilaire: Done. 

I thought this might not work in the case that a different date ever got stored by accident. e.g. via console or database manipulation. 

But this can't happen. The 2000-1-1 date is never stored in the table. It gets added when it's turned into a ruby object.

```
\d users
...
| email_time_of_day      | time without time zone      |                                                                  |
...
```

```
triage_development> select email_time_of_day from users;
+---------------------+
| email_time_of_day   |
|---------------------|
| 15:31:01.986967     |
+---------------------+
```

schneems:
🎉 this is amazing. I want to babysit this deploy a bit, so i'm going to wait until next week to merge. Thanks again for all the work!

elliotthilaire:
> 🎉 this is amazing. I want to babysit this deploy a bit, so i'm going to wait until next week to merge. Thanks again for all the work!

No problem. If you get a PR deploy running I'm happy to test. Let me know if there's anything else that pops up that needs attention before merge.

schneems:
Patch is merged. Thanks again for all your hard work!

Should be deployed shortly.

elliotthilaire:
@schneems Is everything ok with this?

I haven't received an email for the last couple of days.

I've set my preference to be 6am UTC.  The last daily email I received was on the 21st at the unchanged time (5:40~5:50)

schneems:
I think I ran out of sendgrid credits

elliotthilaire:
Ok. That's fine. I was concerned I broke something. If there's an issue with the feature; I'll fix it.

schneems:
Also I forgot to set the scheduler to run hourly instead of once a day. I just fixed that.

schneems:
There's a bug somewhere. I got this report. I'm trmporarilly disabling emails https://twitter.com/rx14_chibi/status/878715739011776512

elliotthilaire:
@schneems Sorry about this. I'm looking into it.

From the screenshot. It's not sending emails out every hour. It send out 7 consecutive emails.
I think `user.days_since_last_clicked` is working on a different day to `before_email_time_of_day?` so there is an overlap.

`days_since_last_clicked` is using Time.now (which returns the time in timezone of the server)
The new feature uses `Time.current` (which returns the time in UTC)

Could you confirm what timezone the server is at? I assume it is about a 7 hour difference to UTC?

I think we can fix this by changing user to use `Time.current` instead of `Time.now` so it it timezone agnostic.

I think this is also related to the random test failure https://github.com/codetriage/codetriage/issues/541

PR #532, opened 2017-02-21
Update readme to prefer `db:schema:load` over `db:migrate`

elliotthilaire:
Follow on from:
https://github.com/codetriage/codetriage/pull/526


schneems:
thanks!

PR #536, opened 2017-03-08
Removed the trailing double quote mark present in the open file path in doc_methods show page in edit section of the page.

ashishwadekar:
Fix for removing double quote from open path in edit section of docs methods

codetriage open path issue


schneems:
Awesome, thank you!

PR #538, opened 2017-04-01
Added documentation for doc_mailer_maker.  Hoping for feedback.

hschreck:


nateberkopec:
What is this style called? AFAICT CodeTriage has no doc style, so merging this would be "adopting" one.

schneems:
I like YARD for doc syntax, though just having comments for internal docs is already better than not having them at all. I'm 👍

PR #546, opened 2017-05-26
add jquery-ujs back

spenceralan:
This reverts commit 89084c1d2c80ab642d13b3b039c2a77e5c5fdb6d,
which had broken the data-method='delete' attribute on the Sign Out link,
thereby causing it to match users#show with an :id of 'sign_out'.


schneems:
thanks

PR #548, opened 2017-05-27
Fix failing test, test speed, clean output

nateberkopec:


test/functional/pages_controller_test.rb
```diff
@@ -22,8 +22,8 @@ class PagesControllerTest < ActionController::TestCase
 
   test "cache control for headers" do
     get :index
-    assert_equal("no-cache, no-store, max-age=0, must-revalidate", @response.headers["Cache-Control"])
-    assert_equal("no-cache", @response.headers["Pragma"])
-    assert_equal("Fri, 01 Jan 1990 00:00:00 GMT", @response.headers["Expires"])
+    assert_equal("no-cache, no-store, max-age=0, must-revalidate", response.headers["Cache-Control"])
```
nateberkopec: no idea why but using the instance variable here results in a 4 second long test

nateberkopec: nvm i think thats just the capybara starting up

PR #554, opened 2017-05-28
Refactor workers

nateberkopec:
+ Prefer ActiveJob serialization
+ Move code out of models
+ Use Minitest mocks
+ Aggressive find_each batch sizes

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,0 +1,3 @@
+class ApplicationJob < ActiveJob::Base
+  queue_as :default
+end
--- a/app/jobs/background_inactive_email_job.rb
+++ b/app/jobs/background_inactive_email_job.rb
@@ -1,8 +1,5 @@
-class BackgroundInactiveEmailJob < ActiveJob::Base
-  queue_as :default
-
-  def perform(id)
-    user = User.find(id.to_i)
+class BackgroundInactiveEmailJob < ApplicationJob
+  def perform(user)
     return false if user.repo_subscriptions.present?
     UserMailer.poke_inactive(user).deliver_now
   end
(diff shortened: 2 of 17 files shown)


nateberkopec:
mfw I only wanted to determine what EmailDecider was for and then did this instead

yakshaving_thumbnail

schneems:
Re-ran heroku tests and they're passing.

Thanks! ❤️

nateberkopec:
I want to figure out how to make all the stuff in the Rake tasks not actually look up and instantiate an AR object.

schneems:
Seems like a legit ability. I retweeted your glibal ID tweet. Not sure if there is an existing way.

At worst can we do a select with only the fields needed to build global ID?

nateberkopec:
actually getting the GID is easy I realize now. but I'm not sure you can just enqueue a job with a string arg that looks like a GID, I think it has to be an object which mixes in GlobalID::Identification

schneems:
Nate: Unexpected problem with this migration. I already had a bunch of jobs enqueued with integers, that all started failing because now we're calling methods on integers since we're expecting them to be objects.

```
NoMethodError: undefined method `populate_docs!' for 2306:Integer
```

Also it looks like several things are untested. Badge generation failed after my 5.1 uprade since `render text:` was changed to `render plain:`. 

This was the other one:

```
NameError: undefined local variable or method `logger' for #<DocMailerMaker:0x0000000b730fe0>
1
File "/app/app/models/doc_mailer_maker.rb" line 59 in deliver
2
File "/app/app/models/user.rb" line 30 in subscribe_docs!
3
File "/app/app/jobs/subscribe_user_to_docs.rb" line 5 in perform
+ 80 non-project frames
```

So it looks like we're not testing the case where a doc is filtered or there are no docs to deliver. 

All in all, not bad. I had to upgrade my rollbar instance to handle all the errors, but after I did that and made a few patches I think we're back in business

Issue #558, opened 2017-06-02
/bin/sh: jemalloc.sh: command not found


elliotthilaire:
https://github.com/codetriage/codetriage/commit/dc26e71601a8887d2b9b01d5841453f09238d597 has introduced a minor documentation bug.

README.md says to launch app with `heroku local` but does not mention you need jemalloc installed.

```
➜  codetriage git:(add-email-time-of-day-preference) heroku local
[OKAY] Loaded ENV .env File as KEY=VALUE Format
5:30:47 PM web.1    |  /bin/sh: jemalloc.sh: command not found
5:30:47 PM worker.1 |  /bin/sh: jemalloc.sh: command not found
[DONE] Killing all processes with signal  null
5:30:47 PM worker.1 Exited with exit code 127
5:30:47 PM web.1    Exited with exit code 127
```

README.md should mention jemalloc is needed or specify a different command to launch the dev server. Not sure which is preferred.

nateberkopec:
yeah sorry about that, adding that to test it on production. I think the best way would probably be a "Procfile.development".

elliotthilaire:
Not a problem. It didn't take long to work out. I got around it with `brew install jemalloc`
I can look at creating a PR for Procfile.development later today. But happy for someone to beat me to it.

nateberkopec:
Nice!

PR #559, opened 2017-06-02
Add procfile development file

elliotthilaire:
Add a Procfile.development file and update README.md on how to use it.

Closes https://github.com/codetriage/codetriage/issues/558

### DIFF
--- a/Procfile.development
+++ b/Procfile.development
@@ -1,0 +1,3 @@
+web: bin/puma -C config/puma.rb
+worker: bundle exec sidekiq -q default -q mailers -c ${SIDEKIQ_CONCURRENCY:-5}
+release: bundle exec rake db:migrate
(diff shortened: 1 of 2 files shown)


nateberkopec:
>I don't think we can patch the jemalloc buildpack to not need to call this file.

thats a lot of double negatives, what?

schneems:
I need to schedule some time to look at the jemalloc buildpack. I think I can have it just work ™️ without any extra LD_PRELOAD or extra script calls, but i'm not sure. This is more of a reminder to myself.

I actually didn't notice the jemalloc change go through. Has it helped metrics?

nateberkopec:
Nah, pretty much left it the same. I haven't dug into that though, because you have puma-worker-killer on and I'm not sure what it's doing. 

Also codetriage's memory usage is pretty good to begin with.

PR #560, opened 2017-06-04
[close #557] Harden Token Check

### LINKED ISSUE #557: Receiving email in error
(Email instructed me to do this!) I received an email about my github account not working, then was able to log into codetriage perfectly fine through my GitHub account

schneems:
- Exit early if API is reported as being down
- Check same token multiple times, only remove if it returns `false` 3 times in a row.
- Archive token in `old_token` instead of just deleting, allows for recovery in case of another mass token invalidation event. 

We should enhance the RequestError from GitHubBub so that it has the response object in it, so we can detect the status code. If GitHub returns anything other than a 404 for that endpoint it means such as a 500 it means we should not delete the endpoint.

### DIFF
--- a/lib/tasks/schedule.rake
+++ b/lib/tasks/schedule.rake
@@ -27,8 +27,20 @@ namespace :schedule do
   end
 
   task check_user_auth: :environment do
+    # API may be down
+    response          = Excon.get("https://status.github.com/api/status.json").body
+    github_api_status = JSON.parse(response)["status"]
+    next unless github_api_status == "good"
+
     User.where.not(token: nil).find_each(batch_size: 100) do |user|
-      user.update_attributes(token: nil) unless user.auth_is_valid?
+      # Check multiple times if token is not valid
+      # if token is valid once, go to next user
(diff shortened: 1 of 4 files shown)


lib/tasks/schedule.rake
```diff
@@ -27,8 +27,20 @@ namespace :schedule do
...
+    # API may be down
+    response          = Excon.get("https://status.github.com/api/status.json").body
+    github_api_status = JSON.parse(response)["status"]
+    next unless github_api_status == "good"
+
     User.where.not(token: nil).find_each(batch_size: 100) do |user|
-      user.update_attributes(token: nil) unless user.auth_is_valid?
+      # Check multiple times if token is not valid
+      # if token is valid once, go to next user
+      next if user.auth_is_valid?
```
nateberkopec: `3.times do { }`?

schneems: The next would escape from that context and not break the outer loop.

nateberkopec: derp

PR #563, opened 2017-06-18
Couple of seeding niceties

nateberkopec:


db/seeds.rb
```diff
@@ -7,14 +7,16 @@
 #   Mayor.create(name: 'Emanuel', city: cities.first)
 require 'faker'
 
+raise "You do not have a Redis server running" unless Sidekiq.redis { |r| r.echo "Alive" }
```
nateberkopec: For some reason, Repos fail to save if there's no Redis server running, despite the test queue adapter.

app.json
```diff
@@ -4,7 +4,8 @@
   "website": "https://www.codetriage.com/",
   "repository": "https://github.com/codetriage/codetriage",
   "scripts":{
-    "postdeploy":"rake db:schema:load db:migrate db:seed"
+    "postdeploy":"DISABLE_DATABASE_ENVIRONMENT_CHECK=1
+ rake db:schema:load db:migrate db:seed"
```
schneems: This shouldn't be required. Are you getting errors?

I wrote that feature. When the database is empty the environment check should basically not care what you do to it.

PR #568, opened 2017-06-23
Check for `before_email_time_of_day?` first

elliotthilaire:
I think this check should be the cheapest.
If it returns true then the other queries do not need to run.

### DIFF
--- a/app/jobs/send_daily_triage_email_job.rb
+++ b/app/jobs/send_daily_triage_email_job.rb
@@ -1,6 +1,6 @@
 class SendDailyTriageEmailJob < ApplicationJob
   def perform(user)
-    return false if !user.repo_subscriptions.any? || skip_daily_email?(user) || before_email_time_of_day?(user)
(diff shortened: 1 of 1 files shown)


elliotthilaire:
This occurred to me:
I have been getting my email around 5:40pm UTC. So I assume that it was scheduled to run at 5:00pm UTC and taking at least 40 minutes to run?

I don't know what would happen if the job takes longer than 1 hour to run when the next run is triggered.

This is a blind optimization that should prevent `skip_daily_email?(user)` and `EmailDecider` called every hour for each user.

schneems:
Sorry, just getting to this now. I made this change in #572.

Digging into the code i'm wondering an issue with scheduler time. The scheduler isn't guaranteed to go off exactly at the hour.

It could fire at 10:59 or 11:01. As written, if you've said you want emails at 11 and scheduler runs at 10:59 it will not send you an email, then you have to wait until 12 (an extra hour) to get your email. Maybe we could see if the time difference between your desired time is less than 60 minutes and then send?

PR #569, opened 2017-06-25
Fix sending duplicate emails

elliotthilaire:
This checks if any of user.repo_subscriptions has a last_sent_at that is set to within the last 24 hours. If that's the case then it will abort the job.


schneems:
Thanks! Is there anything else you want to check or verify before I re-enable emails?

elliotthilaire:
I'm not aware of anything else.

It's hard to know with 100% certainty without testing on a staging site that's configured the same and running jobs.

(This slipped through in the original PR because I thought it was the responsibility of EmailDecider.
I didn't notice the difference between `last_clicked_at` and `last_sent_at`.)

(I'm online for the next couple of hours)

schneems:
I deployed, and i've re-added the task. Thanks again for all the work!

PR #570, opened 2017-06-28
Instrument metrics with statsd

schneems:


nateberkopec:
CoooOOll will this just be sent a 3rd-party Heroku addon or smth?

schneems:
This is part of the language metrics feature I was asking you about a long time ago.

PR #574, opened 2017-06-29
Don't count things we don't have to

schneems:


nateberkopec:
I looked at this a bunch, I wonder why will_paginate doesn't just pick that up automatically.

schneems:
idK could be a good PR to kaminari or will paginate

schneems:
Turns out this PR did nothing. That Option in will paginate does nothing

grillermo:
This would be really useful i get double queries and double response time due to this to the point that i disable the counter in some pretty heavily requested views.

nateberkopec:
I feel like there's something we don't know about pagination and AR, because kaminari has weird logic when it comes to this as well. :|

schneems:
I can switch to "next page" without the numbers style of pagination to avoid the count. I'm not sure if I can pass in total count https://github.com/kaminari/kaminari/pull/681#issuecomment-258727907

Or I could replace with another totally different pagination query scheme.

PR #586, opened 2017-07-07
Add own app to readme

DavidRagone:
It's quite possible I missed something (I see many forks, so seems odd I'd be the first to have this issue). But I was unable to OAuth locally without first creating my own app on GitHub (the default keys suggested above in the README resulted in an error). It's quite possible I misread the README, but I thought this might be helpful. Happy to make any other adjustments if this is useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ public/assets
 
 # ignore vendor/bundle
 vendor/bundle
+tags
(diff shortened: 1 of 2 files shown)


DavidRagone:
😢  that I can't see the details of the Heroku CI test failure.

nateberkopec:
Thanks!

PR #589, opened 2017-07-07
Only show repos I own by default

DavidRagone:
The use of the `active` class on all 3 panes means that "owned",
"starred", and "watched" repos are all displayed, despite the fact that
the UI only has the "owned" tab as active. This is misleading/confusing.
While I think there's some benefit to an "all" option (scrolling is
easier than clicking) and therefore it's possible that this change
results in a decrease in repo added, I think it's worth the improved UX
(and the decreased metric can likely be remedied by ading and defaulting
to an "all repos" tab).

It appears this bug was introduced way back in
324c1a0358f18eba7ce75eee9fb2406baac25f9a when the feature to add starred
and watched repos was introduced.

### DIFF
--- a/app/views/repos/new.html.slim
+++ b/app/views/repos/new.html.slim
@@ -25,10 +25,10 @@
         .tab-pane.active#own
           = render 'shared/add_repos_list', repos: @own_repos
 
-        .tab-pane.active#starred
+        .tab-pane#starred
           = render 'shared/add_repos_list', repos: @starred_repos
 
-        .tab-pane.active#watched
+        .tab-pane#watched
           = render 'shared/add_repos_list', repos: @watched_repos
 
   section.content-section


nateberkopec:
Really want to see this on a review app, but that isn't working r/n. Delegating @schneems

DavidRagone:
Makes sense. In the mean time, hopefully some screenshots will help (apologies for not including those before).

## When loading the /repos/new (hard refresh)


### Before:
localhost-3000-repos-new ipad pro 2

### After:
localhost-3000-repos-new ipad pro

nateberkopec:
Definitely makes sense to me but I'll let @schneems merge, I don't know enough about the purpose of this page.

schneems:
Thanks for the PR. merging in!

>  Really want to see this on a review app, but that isn't working r/n. Delegating @schneems

Looks like there was an issue with the post deploy 

```
I, [2017-07-07T12:52:29.352033 #7]  INFO -- : Loaded Configuration: /app/config/scout_apm.yml. Using environment: production
[Scout] [07/07/17 12:52:29 +0000 run.3351 (7)] INFO : Attempting to start Scout Agent [2.0.0.pre3] on [run.3351]
[Scout] [07/07/17 12:52:29 +0000 run.3351 (7)] WARN : Monitoring isn't enabled for the [production] environment. Not starting agent
Raven 2.5.3 configured not to capture errors: DSN not set
Raven 2.5.3 configured not to capture errors: DSN not set
  [1m[35m (4.2ms)[0m  [1m[34mSELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC[0m
  [1m[35m (8.3ms)[0m  [1m[34mSELECT "ar_internal_metadata"."value" FROM "ar_internal_metadata" WHERE "ar_internal_metadata"."key" = $1[0m  [["key", "environment"]]
  [1m[35m (8.7ms)[0m  [1m[34mSELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC[0m
  [1m[35m (0.8ms)[0m  [1m[34mSELECT "ar_internal_metadata"."value" FROM "ar_internal_metadata" WHERE "ar_internal_metadata"."key" = $1[0m  [["key", "environment"]]
  [1m[35m (1.5ms)[0m  [1m[34mSELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC[0m
  [1m[35m (2.8ms)[0m  [1m[34mSELECT "ar_internal_metadata"."value" FROM "ar_internal_metadata" WHERE "ar_internal_metadata"."key" = $1[0m  [["key", "environment"]]
You are attempting to run a destructive action against your 'production' database.
If you are sure you want to continue, run the same command with the environment variable:
DISABLE_DATABASE_ENVIRONMENT_CHECK=1 excluded from capture: DSN not set
rake aborted!
ActiveRecord::ProtectedEnvironmentError: You are attempting to run a destructive action against your 'production' database.
If you are sure you want to continue, run the same command with the environment variable:
DISABLE_DATABASE_ENVIRONMENT_CHECK=1
/app/vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/tasks/database_tasks.rb:59:in `check_protected_environments!'
/app/vendor/bundle/ruby/2.4.0/gems/activerecord-5.1.1/lib/active_record/railties/databases.rake:11:in `block (2 levels) in <main>'
Tasks: TOP => db:schema:load => db:check_protected_environments
(See full trace by running task with --trace)
```

Which seems odd considering it should be a brand new database. I'm not sure. That task should skip when the DB is empty.

https://github.com/codetriage/codetriage/blob/master/app.json#L7

On a side note, this is now the slowest page on the site. It is slow since we have to get this info from GitHub's API. 

One Idea I had was to use the something like render async https://semaphoreci.com/blog/2017/06/08/speeding-up-rails-pages-with-render-async.html to render the content so the page loads quickly, but each of those tabs can be loaded separately and concurrently.

Also maybe move the "full repo url" box to the top.

DavidRagone:
Love the idea of the async load; given the page defaults to "my repos", it'd be worth even seeing the impact of loading that one synchronously, with the other two async.

PR #590, opened 2017-07-07
Add a facade for all Github HTTP interactions

DavidRagone:
Resolves codetriage#578

I may have gone overboard in introducing distinct GithubFetchers for the different resources, and I definitely went underboard in writing tests. I spent a couple hours at the onset working on testing the existing `populate_docs!` and related methods, but failed miserably (I kept having issues w/ BadRequests to GitHub in the tests), so I continued on the refactor and then tested manually, and 🤞 that the existing coverage will do. I doubt that's true, and am happy to take another crack at tests, though I'd love some guidance or even a remote pair for a little while to get me going in the right direction. 

For now, I'm opening up this PR to get feedback on the approach and style. I'd love to hear if this is along the lines of what @nateberkopec was thinking in codetriage#578


app/controllers/repos_controller.rb
```diff
@@ -34,8 +34,7 @@ def create
       SendSingleTriageEmailJob.perform_later(@repo_sub.id)
       redirect_to @repo
     else
-      response = GitHubBub.get("/user/repos", type: "owner", token: current_user.token)
-      @own_repos = response.json_body
+      @own_repos = GithubFetcher::Repo.repos_for(current_user.token, 'repos', type: 'owner')
```
nateberkopec: I think in my head I was imagining that the github fetcher would be composed:

```ruby
@own_repos = current_user.own_repositories
# meanwhile in User
def own_repositories
  GithubFetcher::Repo.repos_for(current_user.token, 'repos', type: 'owner')
end
```

app/controllers/repos_controller.rb
```diff
@@ -85,10 +84,9 @@ def params_blank?
       repo_params.blank?
     end
 
-    def cached_repos(type, options = {})
-      Rails.cache.fetch("user/#{type}/#{current_user.id}", expires_in: 30.minutes) do
-        repos_response = GitHubBub.get("/user/#{type}", { token: current_user.token }.merge(options))
-        SortedRepoCollection.new(repos_response.json_body)
+    def cached_repos(kind, options = {})
+      Rails.cache.fetch("user/#{kind}/#{current_user.id}", expires_in: 30.minutes) do
+        SortedRepoCollection.new(GithubFetcher::Repo.repos_for(current_user.token, kind, options))
```
nateberkopec: Same thing here. Basically I think GithubFetcher::Repo should only be accessed directly by a model, which is composing it.

app/jobs/populate_issues_job.rb
```diff
@@ -1,9 +1,5 @@
 class PopulateIssuesJob < ApplicationJob
   def perform(repo)
-    begin
-      repo.populate_multi_issues!(state: 'open')
-    rescue GitHubBub::RequestError => e
-      repo.update_attributes(github_error_msg: e.message)
-    end
+    PopulateIssues.(repo)
```
nateberkopec: typo

app/models/github_fetcher/issue.rb
```diff
@@ -0,0 +1,18 @@
+module GithubFetcher
+  class Issue
+    attr_reader :api_path
+
+    def initialize(issue)
+      @api_path = "/repos/#{issue.owner_name}/#{issue.repo_name}/issues/#{issue.number}"
+    end
+
+    def issue_json
```
nateberkopec: maybe call this "as_json"

app/models/github_fetcher/issue.rb
```diff
@@ -0,0 +1,18 @@
...
+      GitHubBub.get(api_path).json_body
+    end
+
+    def commenters
```
nateberkopec: likewise. `commenters_as_json`.

app/models/github_fetcher/issues_response.rb
```diff
@@ -0,0 +1,21 @@
+module GithubFetcher
+  class IssuesResponse
```
nateberkopec: I would put this class back into `Issues`, then give `Issues` a method like:

```ruby
def response
  @response ||= GitHubBub.get(api_path)
end
```

app/models/github_fetcher/null_issue_response.rb
```diff
@@ -0,0 +1,15 @@
+module GithubFetcher
+  class NullIssueResponse
```
nateberkopec: ditto. this can all live in Issues. if there are no issues, set `response` to an empty array.

app/models/github_fetcher/repo.rb
```diff
@@ -0,0 +1,72 @@
+module GithubFetcher
+  class Repo
+    attr_accessor :repo_path
```
nateberkopec: for consistency, `attr_reader :api_path`

app/models/github_fetcher/repo.rb
```diff
@@ -0,0 +1,72 @@
...
+        nil
+      end
+    end
+
+    def clone
+      `cd #{dir} && git clone #{clone_url} 2>&1`
+      return dir
+    end
+
+    def json
```
nateberkopec: `to_json`

nateberkopec: or `as_json`, whichever you think is appropriate.

app/models/github_fetcher/repo.rb
```diff
@@ -0,0 +1,72 @@
...
+      begin
+        IssuesResponse.new(GitHubBub.get(
+          api_issues_path,
+          state:     state,
+          page:      page,
+          sort:      'comments',
+          direction: 'desc'
+        ))
+      rescue GitHubBub::RequestError => e
+        if repo = ::Repo.where(full_name: @full_name).first
```
nateberkopec: This doesn't feel like  GithubFetcher::Repo's responsibility.

app/models/github_fetcher/null_issue_response.rb
nateberkopec: OK nvm I understand now, I didn't realize "Issue" was for a single Issue, not an array of them.

Maybe we need a class like: 

```ruby
module GithubFetcher
   class Issues
     include Enumerable
   end
end
```

See [here](https://stackoverflow.com/questions/7220634/how-do-i-use-the-enumerable-mixin-in-my-class).

app/models/github_fetcher/repo.rb
```diff
@@ -0,0 +1,72 @@
...
+          repo.update_attributes(github_error_msg: e.message)
+        end
+        NullIssueResponse.new # return empty enumerable interface
+      end
+    end
+
+    def api_issues_path
```
nateberkopec: Private?

app/models/github_fetcher/repo.rb
```diff
@@ -0,0 +1,72 @@
...
+      File.join('repos', @full_name, '/issues')
+    end
+
+    def self.repos_for(token, kind, options)
```
nateberkopec: It feels like instead the caller should create a GithubFetcher::Repo with the correct parameters.

app/models/github_fetcher/user.rb
```diff
@@ -0,0 +1,27 @@
+module GithubFetcher
+  class User
+    attr_reader :token
+
+    def initialize(token)
+      @token = token
+    end
+
+    def json
```
nateberkopec: again, it feels like this should be standardized across all of the classes. Maybe you can inherit, like so:

```ruby
module GithubFetcher
  class Resource
    def initialize(api_path, options = {})
      @api_path = api_path
      @options = options
    end

    def as_json
      response.json_body
    end

    def response
      @response ||= GitHubBub.get(api_path, options)
    end
  end
end

module GithubFetcher
  class User < Resource
    def initialize(api_path, options)
      super
      @api_path = "/user"
    end
  end
end
```

app/models/repo.rb
```diff
@@ -24,18 +23,15 @@ def can_doctor_docs?
     CLASS_FOR_DOC_LANGUAGE[self.language]
   end
 
+  def fetcher
+    @fetcher ||= GithubFetcher::Repo.new(path)
+  end
+
   def populate_docs!
     return unless can_doctor_docs?
+    return unless fetcher.commit_sha
```
nateberkopec: *bathes in the cleanliness*

app/models/repo.rb
```diff
@@ -173,37 +159,12 @@ def self.order_by_subscribers
...
-  end
-
   def update_from_github
-    resp = GitHubBub.get(repo_path).json_body
-    self.language    = resp['language']
-    self.description = resp['description']
-    self.stars_count = resp['stargazers_count']
-    self.save
+    self.update(
+      language: fetcher.json['language'],
```
nateberkopec: It looks like our Fetcher's contract is to always return a "null object" of the same type that it would have received. e.g. `json` here should return an empty hash on request failure.

nateberkopec:
Left a bunch of comments. I think you're missing a GithubFetcher::Response class which all the resource types inherit from.

app/models/github_fetcher/issues_response.rb
nateberkopec: the nice thing about a method like this is that it would be super easy to stub in tests...

DavidRagone:
Thanks for all the detailed feedback! Love it.

app/jobs/populate_issues_job.rb
DavidRagone: what's the typo? I'm not seeing it :/

app/models/github_fetcher/null_issue_response.rb
DavidRagone: Good call. I was thinking of moving in this direction, then apparently forgot :)

app/models/github_fetcher/repo.rb
DavidRagone: at least as of now, it's referenced in [this error message](https://github.com/codetriage/codetriage/pull/590/files#diff-0fa2d88e0cbaacf18f6096a7304430a1R202)

app/models/repo.rb
DavidRagone: Good call. Though I imagine in this case, we don't want to over-write what values might already be there

DavidRagone:
WIP.

TODOs:

- [ ] Create base class for fetchers ([see comment](https://github.com/codetriage/codetriage/pull/590/files#r126142126))
- [ ] Only call `fetcher` from within models
- [ ] Figure out better way of handling `IssueResponse` / `NullIssueResponse` logic. Perhaps enumerable
- [ ] Move "what to do with error" logic out of GithubFetcher::Repo; [see comment](https://github.com/codetriage/codetriage/pull/590/files#r126140193)
- [ ] Have `repos_for` instantiate a new fetcher w/ correct params. May mean moving resulting logic into instance methods ([see comment](https://github.com/codetriage/codetriage/pull/590/files#r126141078))

app/models/github_fetcher/user.rb
DavidRagone: I like this approach, though it means that the caller has to know the details of the GitHub api path, which feels a bit odd. But perhaps I'm thinking about it wrong.

nateberkopec: Sorry, I didn't mean for GithubFetcher::User to have required args for initializing.

```ruby
module GithubFetcher
  class User < Resource
    def initialize(user_name, options = {})
      @api_path = "/user/#{user_name}" #example, i dont know if this is actually appropriate
      @options = {}
    end
  end
end
```

app/jobs/populate_issues_job.rb
nateberkopec: I think you meant .call(repo)

app/services/populate_issues.rb
```diff
@@ -0,0 +1,30 @@
+class PopulateIssues
```
nateberkopec: This code is only used by the PopulateIssuesJob right? As in, it isn't used anywhere else?

If so, I don't see the point of making it a separate class when PopulateIssuesJob just calls PopulateIssues.call. Probably should just move all the code into PopulateIssuesJob.

app/jobs/populate_issues_job.rb
DavidRagone: This is actually one of my favorite little ruby-isms: `Foo.()` will call `Foo.call`. I'd be curious to hear if @schneems uses it. When a class has a single entry point that would otherwise be called `call` or `run`, I often prefer this approach because of how nicely it reads.

app/services/populate_issues.rb
DavidRagone: I was debating that as well. The reason I'll often create a separate service class has to do with wanting to provide greater flexibility for the future. If there's a point in time (even now) when one might want to run this manually in the console, I like having a class I can access and use directly. That being said, I suppose there's no reason to not just do that within `PopulateIssuesJob`.

app/jobs/populate_issues_job.rb
nateberkopec: TIL. I've never seen it.

app/services/populate_issues.rb
nateberkopec: > do with wanting to provide greater flexibility for the future

meh YAGNI. It's not any different than just call PopulateIssuesJob.perform imo.

DavidRagone: Yeah, that's pretty much where I landed while live-streaming my thoughts above :)

app/models/github_fetcher/resource.rb
```diff
@@ -1,6 +1,8 @@
 module GithubFetcher
   # Base class for resources accessible via GitHubBub
   class Resource
```
nateberkopec: Nice, this class looks great now!

DavidRagone: Yup - good call on this one! And I'm learning minitest along the way!

app/models/github_fetcher/issue_comments.rb
```diff
@@ -0,0 +1,19 @@
...
+  class IssueComments < Resource
+    def initialize(owner_name:, repo_name:, number:)
+      @api_path = File.join(
+        'repos',
+        owner_name,
+        repo_name,
+        'issues',
+        number.to_s,
+        'comments'
+      )
```
nateberkopec: me like

test/unit/github_fetcher/issue_comments_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    issue.repo = repos(:sinatra_sinatra) # For some reason, this isn't set :/
+
+    GithubFetcher::IssueComments.new(
+      owner_name: issue.owner_name,
+      repo_name: issue.repo_name,
+      number: issue.number
+    )
+  end
+
+  test "quacks like a GithubFetcher::Resource" do
```
nateberkopec: Meh. Just test that it inherits.

test/unit/github_fetcher/issue_comments_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    fetcher = comments_fetcher
+    GithubFetcher::Resource.instance_methods(false).each do |method|
+      assert fetcher.respond_to? method, "Failed to respond_to? #{method}"
+    end
+    GithubFetcher::Resource.private_instance_methods(false).each do |method|
+      assert fetcher.respond_to? method, "Failed to respond_to? #{method}"
+    end
+  end
+
+  test "#as_json doesn't raise errors" do
```
nateberkopec: i don't like "nothing raised" tests because they're basically no-ops http://www.thagomizer.com/blog/2013/06/11/assert-nothing-raised.html

test/unit/github_fetcher/issue_comments_test.rb
DavidRagone: yeah that sounds more reasonable.

app/models/github_fetcher/issue_comments.rb
DavidRagone: funny how once the pattern exists, I start seeing it everywhere

test/unit/github_fetcher/issue_comments_test.rb
nateberkopec: It looks like in this case that you should just test something in the return value of as_json.

test/unit/github_fetcher/issue_comments_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    VCR.use_cassette "fetch_issue_comments" do
+      assert_nothing_raised { comments_fetcher.as_json }
+    end
+  end
+
+  test "#as_json returns list of comments" do
+    first_comment_url = "https://api.github.com/repos/sinatra/sinatra/issues/comments/5"
+
+    VCR.use_cassette "fetch_issue_comments" do
+      assert_equal comments_fetcher.as_json.first['url'], first_comment_url
```
nateberkopec: ...which you do here, making the "nothing raised" test redundant.

test/unit/github_fetcher/issue_comments_test.rb
DavidRagone: Yeah I go back and forth on that. In this case, part of the purpose of the extraction is to not raise this particular error at least, so it felt worthwhile. But I can remove these if you don't think they add enough value

test/unit/github_fetcher/issue_comments_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  test "#as_json returns [] when error" do
+    GitHubBub.stub(:get, -> (_,_) { raise GitHubBub::RequestError }) do
+      assert_equal comments_fetcher.as_json, {}
+    end
+  end
+
+  test "#commenters returns list of unqiue commenters" do
```
nateberkopec: typo

DavidRagone:
21 commits later... I think this is ready for another look @nateberkopec. Appreciate the feedback along the way. FYI, there's a few todos left (all related to one issue) that I'd like to do as a follow-up PR.

DavidRagone:
ping :)

app/jobs/populate_issues_job.rb
```diff
@@ -1,9 +1,41 @@
...
+    end
+  end
+
+  private
+
+  def fetcher(page)
+    repo.issues_fetcher(page: page, state: state)
+  end
+
+  attr_reader :repo, :state
```
nateberkopec: looks like you doubled up here accidentally

app/models/github_fetcher/resource.rb
```diff
@@ -0,0 +1,51 @@
...
+    attr_reader :error_message
+
+    # Often over-ridden by subclasses in order to set @api_path. When over-ridden,
+    #   it's preferable to call `super` so as to set options as below (and get any
+    #   other changes made to this base call in the future)
+    def initialize(options)
+      @options = options
+    end
+
+    # Generally not over-riden
```
nateberkopec: riden -> ridden

app/models/repo.rb
```diff
@@ -173,37 +183,13 @@ def self.order_by_subscribers
...
-    resp = GitHubBub.get(repo_path).json_body
-    self.language    = resp['language']
-    self.description = resp['description']
-    self.stars_count = resp['stargazers_count']
-    self.save
+    json = fetcher.as_json
+    self.update(
+      language: json.fetch('language', language),
+      description: json.fetch('description', description),
+      stars_count: json.fetch('stargazers_count', stars_count),
```
nateberkopec: Remove comma

DavidRagone:
Updated @nateberkopec. Thanks for bearing with me!

schneems:
Wow, thanks for all your work! This is great!

DavidRagone:
Happy to help :)

nateberkopec:
Thanks so much @DavidRagone, great work you can be proud of here!

app/models/issue.rb
```diff
@@ -15,9 +15,24 @@ def valid_for_user?(user, skip_update = Rails.env.test?)
...
+  def fetcher
+    @fetcher ||= GithubFetcher::Issue.new(
+      owner_name: owner_name,
+      repo_name: repo_name,
+      number: number,
+    )
+  end
+
+  def comments_fetcher
+    @fetcher ||= GithubFetcher::IssueComments.new(
```
clembradley: s/@fetcher/@comments_fetcher

schneems: Good catch, can you send a quick PR?

clembradley: No problem. And apologies for the random comment—I happened to come across this PR in my twitter feed when @nateberkopec. Cool refactor!

DavidRagone: oh good find @clembradley, thanks!

Issue #588, opened 2017-07-07
Flaky test in email_rate_limit_test.rb


DavidRagone:
[this spec](https://github.com/codetriage/codetriage/blob/b57ad8fac2fce0e920e3376693b106e8931de76d/test/unit/email_rate_limit_test.rb#L18) would maybe ~1/4 of the time fail locally for me. It's not clear to me why, but seems worth looking into.

DavidRagone:
I can't access the heroku pipeline, but this appears to a spec that is failing on master in CI, intermittently (though I can't force rebuilds on Travis either, so I'm not certain on CI either)

nateberkopec:
Heroku CI is borked r/n because of a bug w/postgres 9.3. I asked @schneems to upgrade but he's out-of-pocket r/n. Watch the Travis builds instead.

schneems:
One of the reasons it is failing is due to https://devcenter.heroku.com/articles/heroku-ci#rails-fixtures-and-referential-integrity. What is the bug with 9.3?

nateberkopec:
@schneems IIRC the fix on Rails only works with Postgres 9.4+

nateberkopec:
so it's strictly inaccurate of me to say "bug with postgres 9.3", but you know what I mean now

lostapathy:
I believe #633 resolved the underlying issue and this can be closed?

prathamesh-sonpatki:
:+1:

PR #591, opened 2017-07-07
Fix flaky test in email_rate_limit_test.rb

DavidRagone:
Resolves https://github.com/codetriage/codetriage/issues/588

When the `clicked_ago` value was divisible by 7, the "once_a_week" spec
failed. It's not clear to me yet if this is an actual bug, or just a bug
in the spec, but I'm assuming the latter and just removing such values
from the sample for now

Also breaks lines in the 80-100 range because I'm a vim/tmux person and working on a 13" mac air :)

### DIFF
--- a/test/unit/email_rate_limit_test.rb
+++ b/test/unit/email_rate_limit_test.rb
@@ -12,10 +12,12 @@ class EmailRateLimitTest < ActiveSupport::TestCase
     valid_values          = seed_array.map { |n| n * multiplier }
     invalid_values        = last_clicked_days_ago.to_a - valid_values
     day_ago               = rand(last_clicked_days_ago)
-    clicked_ago           = valid_values.sample
-    assert EmailRateLimit.new(day_ago).now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
+    clicked_ago           = valid_values.reject { |int| (int % 7).zero? }.sample
+    assert EmailRateLimit.new(day_ago).now?(clicked_ago),
+      "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
     # Use user email frequency settings
-    assert_not EmailRateLimit.new(day_ago, minimum_frequency: "once_a_week").now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}, minimum_frequency: 'once_a_week').now?(#{clicked_ago}) to be false, was not"
+    assert_not EmailRateLimit.new(day_ago, minimum_frequency: "once_a_week").now?(clicked_ago),
+      "Expected  EmailRateLimit.new(#{day_ago}, minimum_frequency: 'once_a_week').now?(#{clicked_ago}) to be false, was not"
   end
 
   test "wait" do


nateberkopec:
my face when I wake up to see all of these PRs

nateberkopec:
> It's not clear to me yet if this is an actual bug, or just a bug
in the spec,

I think this is a bug

nateberkopec:
whoops. anyway, I think we should fix the bug here, not the test.

DavidRagone:
I spent some more time on this, which ended up as a value object extraction. I'll put a PR up for it. 

But I'm still not seeing a bug. If a user says "I want emails once/week" and we last sent an email 7 days ago, why wouldn't we send one now? I'm not sure it matters that they clicked recently, they still got an email a week ago, and it's been a week.

test/unit/email_rate_limit_test.rb
```diff
@@ -12,10 +12,12 @@ def seed_array
...
     day_ago               = rand(last_clicked_days_ago)
-    clicked_ago           = valid_values.sample
-    assert EmailRateLimit.new(day_ago).now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
+    clicked_ago           = valid_values.reject { |int| (int % 7).zero? }.sample
+    assert EmailRateLimit.new(day_ago).now?(clicked_ago),
+      "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
     # Use user email frequency settings
-    assert_not EmailRateLimit.new(day_ago, minimum_frequency: "once_a_week").now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}, minimum_frequency: 'once_a_week').now?(#{clicked_ago}) to be false, was not"
+    assert_not EmailRateLimit.new(day_ago, minimum_frequency: "once_a_week").now?(clicked_ago),
+      "Expected  EmailRateLimit.new(#{day_ago}, minimum_frequency: 'once_a_week').now?(#{clicked_ago}) to be false, was not"
```
DavidRagone: I'm honestly not sure why this expectation is here :/

DavidRagone: @schneems any thoughts on this one?

DavidRagone:
ping :)

nateberkopec:
Making this richard's problem because I don't really understand this class yet tbh

test/unit/email_rate_limit_test.rb
```diff
@@ -12,10 +12,12 @@ def seed_array
     valid_values          = seed_array.map { |n| n * multiplier }
     invalid_values        = last_clicked_days_ago.to_a - valid_values
     day_ago               = rand(last_clicked_days_ago)
-    clicked_ago           = valid_values.sample
-    assert EmailRateLimit.new(day_ago).now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
+    clicked_ago           = valid_values.reject { |int| (int % 7).zero? }.sample
```
schneems: Can you walk me through why this helps? This is in the "daily email frequency" section and not in the "once a week" section.

test/unit/email_rate_limit_test.rb
schneems: When a user sets their email settings to "once_a_week" then we want to make sure that they don't get an email for any frequency lower, which in this case would be daily.

It's testing that the logic provided by the `minimum_frequency:` option works.

lostapathy:
I believe this can be closed now that #633 has been merged.

PR #592, opened 2017-07-07
Extract EmailFrequency Value Object

DavidRagone:
Represents the frequency with which email should be sent. Goal is to simplify the logic for "should we send an email now?", which compares user preference with user actions

In trying to understand the logic and tests in https://github.com/codetriage/codetriage/pull/591, I started moving some code around and ended up with this value object extraction. Not sure if it's wanted, but figured I'd share it nonetheless.

### DIFF
--- a/app/models/email_frequency.rb
+++ b/app/models/email_frequency.rb
@@ -1,0 +1,67 @@
+# Value Object representing the frequency with which email is sent. Used to
+#   compare two different frequencies, e.g. between a set value and calculated
+#   value.
+class EmailFrequency
+  include Comparable
+
+  # More frequent has a higher index, less frequent a lower index. The index is
+  #   used to compare the frequencies.
+  FREQUENCIES = [
+    WAIT         = "wait",
+    ONCE_A_MONTH = "once_a_month",
+    ONCE_A_WEEK  = "once_a_week",
+    TWICE_A_WEEK = "twice_a_week",
+    DAILY        = "daily"
+  ]
+
+  USER_FREQUENCIES = FREQUENCIES - [WAIT]
+
+  # Human-readable string representing the frequency; based on the constant
+  #   FREQUENCIES above.
+  attr_reader :frequency_description
+
+  def initialize(frequency_description)
+    raise ArgumentError unless frequency_description.in? FREQUENCIES
+    @frequency_description = frequency_description
+  end
+
+  # Constructor function that turns an int (num_days) into an EmailFrequency
+  def self.from_days(num_days)
+    case num_days
+    when 0..3
+      new(DAILY)
+    when 7..13
+      new(TWICE_A_WEEK)
+    when 14..30
+      new(ONCE_A_WEEK)
+    when 31..Float::INFINITY
+      new(ONCE_A_MONTH)
+    else
+      new(WAIT)
+    end
+  end
+
+  # With the Comparable module, this allows for `>` and `<` comparisons
+  def <=>(other)
+    FREQUENCIES.index(frequency_description) <=>
+      FREQUENCIES.index(other.frequency_description)
+  end
+
+  def eql?(other)
+    frequency_description == other.frequency_description
+  end
+
+  # Alias of `>`, used for easier code comprehension
+  def more_often?(other)
+    self > other
+  end
+
+  # Alias of `<`, used for easier code comprehension
+  def less_often?(other)
+    self < other
+  end
+
+  def to_s
+    frequency_description
+  end
+end
--- a/app/models/email_frequency.rb
+++ b/app/models/email_frequency.rb
@@ -52,16 +52,6 @@ class EmailFrequency
     frequency_description == other.frequency_description
   end
 
-  # Alias of `>`, used for easier code comprehension
-  def more_often?(other)
-    self > other
-  end
-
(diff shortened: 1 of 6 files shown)


app/models/email_rate_limit.rb
```diff
@@ -39,29 +35,7 @@ def now?(last_sent_days_ago)
...
-        "twice_a_week"
-      when 14..30
-        "once_a_week"
-      when 31..Float::INFINITY
-        "once_a_month"
-      else
-        "wait"
-      end
+    def less_frequent_of_activity_and_preference
+      [@frequency_from_last_click, @frequency_from_preference].min
```
DavidRagone: My goal in the refactor was to simplify the logic in these two private methods.

test/unit/email_rate_limit_test.rb
```diff
@@ -2,7 +2,8 @@
 
 class EmailRateLimitTest < ActiveSupport::TestCase
   def seed_array
-    @seed_array ||= (1..350).to_a # max value has to be divisible by the multiplier of th higest used value i.e. 10_000/30.0
+    # max value has to be divisible by the multiplier of th higest used value i.e. 10_000/30.0
```
DavidRagone: This file just has line length changes (apologies for being difficult like that)

app/models/email_frequency.rb
```diff
@@ -0,0 +1,67 @@
+# Value Object representing the frequency with which email is sent. Used to
+#   compare two different frequencies, e.g. between a set value and calculated
+#   value.
+class EmailFrequency
+  include Comparable
+
+  # More frequent has a higher index, less frequent a lower index. The index is
+  #   used to compare the frequencies.
+  FREQUENCIES = [
+    WAIT         = "wait",
```
nateberkopec: maybe leave a comment explaining what this particular one means, not immediately obvious.

nateberkopec:
👍 just wanted an extra comment in there re: "WAIT" and why it's special

test/unit/email_rate_limit_test.rb
```diff
@@ -51,8 +58,10 @@ def seed_array
     day_ago               = rand(last_clicked_days_ago)
     clicked_ago           = valid_values.sample
     bad_clicked_ago        = invalid_values.sample
-    assert EmailRateLimit.new(day_ago).now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
-    assert EmailRateLimit.new(day_ago).skip?(bad_clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}).skip?(#{bad_clicked_ago}) to be true, was not"
+    assert EmailRateLimit.new(day_ago).now?(clicked_ago),
```
nateberkopec: Let's fix these assertions. Let's create a new assertion, called `assert_true`, which calls `assert_equal true, argument`.

I understand wanting to avoid assert b/c the error messages are basic, but we can DRY this up.

nateberkopec: or it can call assert, with a custom error message. your call on what looks better.

DavidRagone: Will do. Thanks for the feedback!

test/unit/email_rate_limit_test.rb
```diff
@@ -13,9 +9,9 @@ def seed_array
     invalid_values        = last_clicked_days_ago.to_a - valid_values
     day_ago               = rand(last_clicked_days_ago)
     clicked_ago           = valid_values.sample
-    assert EmailRateLimit.new(day_ago).now?(clicked_ago), "Expected  EmailRateLimit.new(#{day_ago}).now?(#{clicked_ago}) to be true, was not"
+    assert_true { EmailRateLimit.new(day_ago).now?(clicked_ago) }
```
DavidRagone: let me know what you think of this approach, @nateberkopec

nateberkopec: Remind me why this needs to be a block?

DavidRagone: I was looking for a way to dry up the assertions (and specifically the description of each, which effectively duplicate the method call). I'm a MiniTest noob (I've almost exclusively used rspec); one day I'll get around to reading the docs, but in the meantime I was curious of how I could essentially run some code and then, if it fails, print out what it was that failed. Whether using `assert foo?` or `assert_equal true, foo?`, the error messages appear to be pretty useless. By passing the block to the `assert_true` method I wrote (see below for implementation), I can use an error message that includes the contents of the block (so the error message is useful, reading `"Expected EmailRateLimit.new(5).now?(10) to be true"`). If there's a better way to do this, I'm all ears.

nateberkopec: > assert_equal true, foo?, the error messages appear to be pretty useless.

I think if that's what you want, you should just print the whole line, like RSpec does.

> Failure/Error: expect(@breadcrumbs.empty?).to be false
> expected false
> got true

There's probably already a minitest formatter that does this.

schneems: Will this give the numbers in the output? Or will it always show the same string "EmailRateLimit.new(day_ago).now?(clicked_ago)". 

That's why I added those outputs, because otherwise when the tests randomly fail it's impossible to know why.

PR #596, opened 2017-07-26
Fix fetcher memoization bug

clembradley:
Both `Issue#fetcher` and `Issue#comments_fetcher` were using the same instance variable (`@fetcher`) when memoizing.  

Wasn't sure if it was worth adding a test to expose the memoization bug or not. Let me know if you feel strongly about it, and I can add one.


schneems:
Thanks! 🙏 if it's easy to test I would be happy to have one for it.

PR #599, opened 2017-07-26
Fix Github response memoization

nateberkopec:
Changing options blows the cache.
+ Add page change method
+ Move last page? to Resource
+ Remove state setting on job, b/c it's just the default.

### DIFF
--- a/app/models/github_fetcher/resource.rb
+++ b/app/models/github_fetcher/resource.rb
@@ -1,7 +1,7 @@
 module GithubFetcher
   # Base class for resources accessible via GitHubBub
   class Resource
-    attr_reader :error_message
+    attr_reader :error_message, :page
 
     # Often over-ridden by subclasses in order to set @api_path. When over-ridden,
     #   it's preferable to call `super` so as to set options as below (and get any
@@ -15,7 +15,7 @@ module GithubFetcher
       @as_json ||= response.json_body
     end
 
-    # Generally not over-riden
+    # Generally not over-ridden
     def response
       @response ||= begin
                       GitHubBub.get(api_path, options)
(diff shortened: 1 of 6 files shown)


app/models/github_fetcher/resource.rb
```diff
@@ -12,18 +12,24 @@ def initialize(options)
 
     # Generally not over-ridden
     def as_json
-      @as_json ||= response.json_body
+      @as_json ||= Hash.new do |h, key|
+                     h[key] = response.json_body
+                   end
+      @as_json[options]
```
schneems: I think this is going to blow out memory. For kubernetes we will have 176 keys and just as many bodies in the hash (which are not small). Let's not memoize here.

test/jobs/populate_issues_job_test.rb
```diff
@@ -29,11 +29,11 @@ class PopulateIssuesJobTest < ActiveJob::TestCase
 
   test "#populate_multi_issues stores error info when fails" do
     repo = repos(:issue_triage_sandbox)
-    fetcher_double = OpenStruct.new(error?: true, error_message: 'something went wrong')
-
-    repo.stub(:issues_fetcher, -> (_) { fetcher_double }) do
-      PopulateIssuesJob.perform_now(repo)
-      assert_equal repo.github_error_msg, 'something went wrong'
+    def repo.issues_fetcher
+      OpenStruct.new(error?: true, error_message: 'something went wrong', page: 1)
```
schneems: I like this. Who needs mocha 😜

app/models/github_fetcher/resource.rb
```diff
@@ -12,18 +12,24 @@ def initialize(options)
...
-    # Generally not over-riden
+    # Generally not over-ridden
     def response
-      @response ||= begin
-                      GitHubBub.get(api_path, options)
-                    rescue GitHubBub::RequestError => e
-                      @error = e
-                      @error_message = e.message
-                      null_response(e)
+      @response ||= Hash.new do |h, key|
```
schneems: Same with hash memoization here, will blow out memory

nateberkopec:
.@schneems pointed out the response hash will get huge when paginating a few hundred times, not a good idea. need to blow it away completely.

nateberkopec:
Fixed

PR #602, opened 2017-08-01
Make all emails deliver_later

### LINKED ISSUE #579: Make all emails deliver_later
Delivering during the request is for weenies!

### LINKED ISSUE #584: Deal with mailer timeouts & `deliver_now`
https://sentry.io/share/issue/3137333132372e333035323232373539/

igorbozato:
Closes #584, closes #579

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -28,6 +28,12 @@ class ActionDispatch::IntegrationTest
   end
 end
 
+class ActionMailer::MessageDelivery
+  def deliver_later
+    deliver_now
+  end
+end
+
 OmniAuth.config.test_mode = true
 OmniAuth.config.add_mock(:github, {
   uid: 'mockstar',
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -28,12 +28,6 @@ class ActionDispatch::IntegrationTest
   end
 end
 
-class ActionMailer::MessageDelivery
-  def deliver_later
-    deliver_now
-  end
-end
-
 OmniAuth.config.test_mode = true
 OmniAuth.config.add_mock(:github, {
   uid: 'mockstar',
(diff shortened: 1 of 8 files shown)


schneems:
I think all of these places are only called from within background jobs currently.

test/test_helper.rb
```diff
@@ -28,6 +28,12 @@ class ActionDispatch::IntegrationTest
   end
 end
 
+class ActionMailer::MessageDelivery
```
nateberkopec: Is this a common thing? I would think there was a config setting somewhere.

nateberkopec: It would probably be better to change the tests to look for "enqueued" email deliveries, tbh, so that you're actually testing that deliver_later is called.

nateberkopec:
This will fix #584 though, because Sidekiq will just retry mails that timeout.

nateberkopec:
The danger with not using deliver_later in jobs is doing stuff like:

```ruby
def perform
  do_some_non_idempotent_stuff
  mail.deliver_now
end
```

...because if that gets retried, bad things happen.

schneems:
Ahh I bet that's what happened with the guy who got 500 issues.

nateberkopec:
Which job was that @schneems? Certainly could have.

test/test_helper.rb
igorbozato: Agreed, using `assert_enqueued_jobs` for testing emails now.

schneems:
Thanks!

schneems:
Nate: this section of code would be called over and over again, each time adding new issues https://github.com/codetriage/codetriage/blob/master/app/models/user.rb#L152.

We should add some logic to either clear out existing non-delivered issues before assigning, or to skip assignment if non-delivered issues already exist.

PR #603, opened 2017-08-02
Use git checkout in instructions instead of git co

jonmast:
Git doesn't have a `co` command, I think we want `checkout` instead. The original author probably had it as an alias.


prathamesh-sonpatki:
Thanks 👍 and welcome 🎉

PR #607, opened 2017-08-09
Switch to local time 2.0 rather than 2_0 branch

benoittgt:
The [branch 2_0 has been deleted and merged](https://github.com/basecamp/local_time/pull/55#event-1195558086
).
It was not possible to build the projet with the previous gemfile.

Issue #606, opened 2017-08-08
Add rubocop ?


benoittgt:
Hello

Add rubocop to codetriage [has been mentioned on twitter by schneems](https://twitter.com/schneems/status/893488921660858369). As I said I can work on this but need to know before which « cops » we want and also the way you want this to be done.

For the cops what about using the one from rails? https://github.com/rails/rails/blob/master/.rubocop.yml

For the process, I can open a PR for each folder (or an average of 50 offenses fixed)?

Thanks

nateberkopec:
First step is to just add it with `--auto-gen-config` to prevent any regressions further than what we've already got 😆

mikegee:
Since #608 is merged, close this?

schneems:
Yup, thanks

PR #608, opened 2017-08-09
Add initial rubocop config

benoittgt:
Hello

As mention in #606. Here is the initial commit for rubocop. I added the todo list of offenses. We still have to choose which "cops" we enable and sometimes at which level.

This PR rely on #607 otherwise I was not able to build the app.


nateberkopec:
Merci! Does that Travis CI output look right to you? It doesn't look like it's only running on changed files.

schneems:
I'm not an expert here. Tests are passing. Good to merge?

nateberkopec:
Hold up, @benoittgt added something that was supposed to only run rubocop on changed files I think but it doesnt appear to work. Wanna hear his reply first.

I am also OK with just running Travis on all files every build.

benoittgt:
Yes. I don't seem that Travis is impacted about the result of rubocop. I'm gonna have a look.

benoittgt:
Ok it works for Travis. If I run on all files I have two offenses that's why the CI failed.

nateberkopec:
Cool, fix 'em and lets merge this.

benoittgt:
I was able to fix one. But the one in `config/environments/production.rb:89` was a mystery for me.

I tried
```ruby
  config.cache_store = if ENV["MEMCACHIER_SERVERS"]
                         :dalli_store,
                           (ENV["MEMCACHIER_SERVERS"] || "").split(","),
                           {:username => ENV["MEMCACHIER_USERNAME"],
                            :password => ENV["MEMCACHIER_PASSWORD"],
                            :failover => true,
                            :socket_timeout => 1.5,
                            :socket_failure_delay => 0.2
                         }
                       else
                         :dalli_store
                       end
```

But
```
/Users/bti/.rvm/gems/ruby-2.4.1/gems/bootsnap-0.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require': /Users/bti/code/codetriage/config/environments/production.rb:90: syntax error, unexpected ',', expecting keyword_end (SyntaxError)
                 :dalli_store, (ENV["MEMCACHIER_SERVERS"] ||
```

So it's disabled for this part. If you have a solution I will change it.

nateberkopec:
Good enough 👍  Thanks @benoittgt!

PR #610, opened 2017-08-10
First batch of "Fix rubocop offenses"

benoittgt:
Hello

Here is the first batch of style corrections. 

Prefer this link to review changes: https://github.com/codetriage/codetriage/pull/610/files?w=1 
It will no display whitespace changes.

PR #611, opened 2017-08-11
Enable first batch of cops layout and add rubocop task

benoittgt:
Hello

This PR:
* Add rubocop as default rake task with test
* Fix first batch of Layout offenses

Sorry for the noise on #610 and for the second review.

🌝


nateberkopec:
Works for me!

PR #612, opened 2017-08-11
Second batch of Rubocop offense fix

benoittgt:
A little bit more changes on this one than spaces and commas. I need sometimes to rewrite small part. When I did it I put the change into a commit.


app/controllers/pages_controller.rb
```diff
@@ -3,19 +3,19 @@ class PagesController < ApplicationController
 
   def index
     @repos = Repo.with_some_issues
-    if language = valid_params[:language] || current_user.try(:favorite_languages)
+    if (language = valid_params[:language] || current_user.try(:favorite_languages))
```
nateberkopec: I think this should be if `(language = valid_params[:language]) ||`

benoittgt:
Thanks for the review !

schneems:
awesome sauce! Thanks for all the work here ❤️

Issue #615, opened 2017-08-15
Email time of day preference may not be working.


elliotthilaire:
I haven't had this work for me since it was deployed. I still recieve emails at around 6pm UTC.

I think there is a flaw in this PR:
https://github.com/codetriage/codetriage/pull/569

I think `email_sent_in_last_24_hours?` needs to be updated to `email_sent_today?`

Checking for an email send in last 24 hours means that if you change the preference to earlier, the email will still get held back by this until the 24 hours is up, and this cycle will continue.

Sorry for not reporting this sooner.

schneems:
Good catch. I hate the intersection of time and programming.

elliotthilaire:
I'm at a Ruby hack night (https://www.meetup.com/ruby-hacknight-london/) and have been working on this with @charlietarr1.

The night is about wrap up but we have found the issue (and another) and have some failing tests. We can submit a fix in the coming days.

These the tests we were working with in their raw form.

```
  test '' do
    def @user.issue_assignments_to_deliver; IssueAssignment.all.limit(1); end
    def @user.email_time_of_day; Time.utc(2000, 1, 1, 04, 0, 0); end

    Time.stub(:now, time_preference_for_today(@user.email_time_of_day) + 1.hour) do
      assert @job.perform(@user)
    end

    Time.stub(:now, time_preference_for_today(@user.email_time_of_day) + 1.hour + 1.day ) do
      assert @job.perform(@user)
    end
  end

  test 'two' do
    def @user.issue_assignments_to_deliver; IssueAssignment.all.limit(1); end
    def @user.email_time_of_day; Time.utc(2000, 1, 1, 05, 0, 0); end

    Time.stub(:now, time_preference_for_today(@user.email_time_of_day)) do
      assert @job.perform(@user)
    end
  end

  test 'thrree' do
    def @user.issue_assignments_to_deliver; IssueAssignment.all.limit(1); end

    def @user.email_time_of_day; Time.utc(2000, 1, 1, 18, 0, 0); end

    Time.stub(:now, time_preference_for_today(@user.email_time_of_day)) do
      assert @job.perform(@user)
    end

    def @user.email_time_of_day; Time.utc(2000, 1, 1, 04, 0, 0); end

    Time.stub(:now, time_preference_for_today(@user.email_time_of_day) + 1.day) do
      assert @job.perform(@user)
    end
  end

  test 'four' do
    def @user.issue_assignments_to_deliver; IssueAssignment.all.limit(1); end

[...]

PR #616, opened 2017-08-18
Fix email not sending at preferred time

elliotthilaire:
Fixes https://github.com/codetriage/codetriage/issues/615

Checking for a sent email in last 24 hours means that if you change the preference to earlier, the email will still get held back until the 24 hours is up, and this cycle will continue.

Instead we check if an email has been sent since the start of the day.

Thanks to @charlietarr1 for pairing with me on this.


schneems:
Thanks!

PR #617, opened 2017-08-18
Appease rubocop

elliotthilaire:
I know this is only temporary but travis is failing PR's early because of it.


nateberkopec:
Fixed on master, thanks

PR #618, opened 2017-08-18
Rename email rate limiting methods

elliotthilaire:
This is a rename of `skip_daily_email?` method to `emails_rate_limited?`

When worked on this Job in the past I misunderstood why `skip_daily_email?` was skipping emails. 

I thought it did more than skip because of the rate limiter. I introduced a bug (now been fixed) because of that misunderstanding. I think this rename makes it more clear.


benoittgt:
With [the commit](https://github.com/codetriage/codetriage/commit/b6d4dbf233b29ef26a3a617d6e10e647c72676b5) from Nate it should be green

schneems:
I rebased with master which had some changes to that file. Waiting for tests now.

Issue #619, opened 2017-08-18
Rubocop offenses prevent Travis from running tests


elliotthilaire:
As a contributor I want to see the full test output from Travis even if something fails.

With the current config, a Rubocop failure will cause travis to stop early and not run the test suite.

Sometimes a Rubocop offense will make it into master. When this happens we can't see if PR's pass the test suite on Travis.

I was about to create a PR to remove rubocop out of the default rake task and into the travis.yml.
But then noticed that would be undoing the work of this commit https://github.com/codetriage/codetriage/pull/611/commits/0eeb4333325d3baaee0ca2a892e398ad8e8e6676

I have opened this issue instead because I don't know the context. 

What was the reason for putting rubocop in the default rake task? 
And can it we move it back, or find another way to get the output of both tests and rubocop showing in Travis?

@benoittgt

nateberkopec:
TBH I like this behavior.

It's maintainer's fault (cough cough me cough cough @schneems) if stuff that breaks rubocop gets merged into master, not Rubocop's fault. We can fix the human problem here.

schneems:
Master should be green, if something goes in that makes it not green, we should revert.

benoittgt:
Hello @elliotthilaire 

Sorry for this unpleasant behavior. 

**I saw this behavior on puma and implemented the same here.** 🤷‍♂️ 
At the beginning task were splitted (it will still output one result for travis) but travis was running test, then rubocop in separate task. I don't think it's possible to have two results rubocop and travis output and I don't think we want this. 

If you are again one or few cop from Rubocop feel free to open an issue. 😉

elliotthilaire:
Hello.

I have made an example app with 4 branches and set it up with travis to demonstrate how separating the steps in .travis.yml works.

i.e,
```
script:
  - bundle exec rubocop
  - bin/rails test
```

Please take a moment to look at:
https://travis-ci.org/elliotthilaire/rails-travis-example-app/branches

* tests-pass-rubocop-passes (build pass)
* tests-pass-rubocop-fails  (fails build)
* tests-fail-rubocop-passes (fails build)
* tests-fail-rubocop-fails  (fails build)

If you have a look at the build logs you will see output from both tests and rubocop and the build still fails appropriately. 

e.g. the tests-fail-rubocop-fails case: https://travis-ci.org/elliotthilaire/rails-travis-example-app/builds/266084051

[...]

PR #620, opened 2017-08-19
Split test and rubocop tasks for Travis

benoittgt:
As discussed with @elliotthilaire on #619 

> I wasn't suggesting we make the build green if it shouldn't be. Instead if the rubocop and tests fail, we can find out in a single build instead of two.


schneems:
Thanks!

PR #621, opened 2017-08-31
Update GitHubBub to v1.0

schneems:
GitHubBub no longer raises exceptions by default. Also `GitHubBub.valid_token?` now explicitly checks for a 404 response to return a false rather than returning false on anything but 200.

Also adds rate limiting code.

See: https://github.com/schneems/git_hub_bub

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -119,7 +119,7 @@ GEM
       dotenv (= 2.2.1)
       railties (>= 3.2, < 5.2)
     erubi (1.6.0)
-    excon (0.45.4)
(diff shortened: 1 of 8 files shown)


schneems:
Can safely ignore Heroku CI failing for now Travis is green.

Would appreciate some eyes @nateberkopec and @DavidRagone 

I eventually want to change how we're doing token invalidations. Instead of batching them I want to respond to them when we get a 401 response from GitHub. I wanted to do that in a separate PR from this for clarity, but if you see something that could make that move easier in the future let me know.

PR #624, opened 2017-10-08
Remove N+1 on UserMailer#send_daily_triage

esparta:
When UserMailer render the emails, the view have a N+1 since it is trying to get properties from issue, user and repo relations on the loop of assignments.

https://github.com/codetriage/codetriage/blob/b3d7a1186e21608052f48d9a9c86eb4c400b7b40/app/views/user_mailer/send_daily_triage.md.erb#L6-L16

Also:

- removing an extra assignment on the assignments ivar
- moving the set_and_check_user guard before performing the query


prathamesh-sonpatki:
Looks good, thanks 👍

esparta:
You're welcome! Thanks for your time reviewing this.

PR #626, opened 2017-10-10
update README to match my experience setting up

lostapathy:
I setup codetriage locally and updated a few things in the README to match my experience.

* Per [Heroku docs](https://devcenter.heroku.com/articles/heroku-local) default port is 5000
* Updated link to register the app with github and updated the title to match the github page
* removed GITHUB_API_KEY parameter that was not needed during my setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,18 +63,17 @@ See the Download page on Redis.io for steps to install on other systems: [http:/
 
 ### Environment
 
-If you want your users to sign up with Github, create a [GitHub Client Application](https://github.com/settings/applications). The urls you are asked to provide will be something like this:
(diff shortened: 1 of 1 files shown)


nateberkopec:
Let's omit the port change. It doesn't really matter what port Heroku deploys it on, the Rails default is 3000 so let's just stick with that.

lostapathy:
@nateberkopec I understand that's the default port for "rails s", but the README calls for running this via "heroku local", and a default "heroku local" runs it locally on 5000.

Specifically in this section https://devcenter.heroku.com/articles/heroku-local#run-your-app-locally-using-the-heroku-local-command-line-tool-start-your-app-locally

Issue #541, opened 2017-04-26
EmailDeciderTest failure


schneems:
```
EmailDeciderTest#test_decides_email_frequency [/home/travis/build/codetriage/codetriage/test/unit/email_decider_test.rb:16]:
Expected  EmailDecider.new(3, 'once_a_week').now?(238) to be false, was not
```

nateberkopec:
i've seen this randomly too, pretty weird

nateberkopec:
This must be a test ordering thing.

nateberkopec:
Or maybe a legit failure from the random days_ago

schneems:
Or a time zone thing. All the values from those random ranges should be valid.

lostapathy:
Note: this test got renamed to EmailRateLimitTest.

Are you guys still seeing this, or has it gone away?  Do you need to run the full test suite to get it to happen, or will it happen just running that test?

lostapathy:
@nateberkopec  I believe this can be closed now that #633 has been merged.  I guess I should have noted that on the PR so this would have auto-closed.

PR #633, opened 2017-10-17
Fix issue with unreliable test

lostapathy:
This PR is to address #541

This test was computing date ranges for daily notification, but then asserting EmailRateLimit behavior based on a weekly preference.  Whenever the random value for clicked_ago was divisible by 7, the last assertion would fail, but in all other cases it passed.

I rewrote the assertion to pass a daily preference (to match the logic of how clicked_ago was generated), then added similar assertion to other intervals.

I ran this one-liner to confirm before and after behavior.
```
for ((i=1;i<=100;i++)); do rake test TEST=test/unit/email_rate_limit_test.rb ; done;
```


nateberkopec:
Yay! Thanks!

schneems:
Awesome, thanks!

PR #634, opened 2017-10-24
barnes

schneems:


Schwad:
@schneems is the main focus of this issue to include the `barnes` gem and bump to ruby 2.5.0p-1 ?

schneems:
Goal is to include barnes. Updating to 2.5.0 was a mistake. Won't work due to https://github.com/tmm1/stackprof/issues/84

PR #640, opened 2017-11-03
Fixed repo name generation in seed file fixes #639

### LINKED ISSUE #639: ActionController::UrlGenerationError - Due to incorrect repo name generation in seed file
I observed that while locally installing the repo and after executing the command `rake db:seed`,  it gives the below error when trying to access http://localhost:5000/.

screen shot 2017-11-03 at 1 34 47 pm

Checked the source code and found that use of `Faker::Hipster.word` in `db/seeds.rb` for generating repo name is sometime generating words with special characters inside it and thus causing the issue.

anuj-verma:


prathamesh-sonpatki:
Thanks

PR #641, opened 2017-11-08
Show languages list vertically instead of horizontally, closes #632

### LINKED ISSUE #632: Sort languages list vertically instead of horizontally
This is a pretty small user experience thing, but I think it would be better to sort the list of languages vertically instead of horizontally. [This Stack Exchange answer explains why](https://ux.stackexchange.com/questions/17176/alphabetical-lists-horizontal-or-vertical).

TL;DR: it's easier and faster to find stuff when vertically sorted

anuja-joshi:


prathamesh-sonpatki:
@anuja-joshi It looks good but it has one issue: http://take.ms/ih5Hv

I think it is happening because the no. of languages on my machine is less than what CSS can distribute over 4 columns. Is there any alternative for this?

anuja-joshi:
@prathamesh-sonpatki we can add conditional classes to apply `column-count` based on number of entries in list, but I think there is no need to of adding this complexity. If number of languages are more than 10 it is anyway going to be distributed in 4 columns (not leaving any empty space). What do you think?

anuja-joshi:
@prathamesh-sonpatki  I picked this issue up with the consent of @anuj-verma

prathamesh-sonpatki:
Thanks @anuja-joshi @anuj-verma 👍 and welcome to the contributor's list! 🎉

prathamesh-sonpatki:
@anuja-joshi The deployment is failing because of this error: https://gist.github.com/prathamesh-sonpatki/e9ec3933f5b3f96acfc6d4d63a9ae811 Have you come across this? cc @schneems

anuj-verma:
@prathamesh-sonpatki, Looked into this and found this answer:
https://stackoverflow.com/questions/32209338/fails-to-push-to-heroku-sprocketsfilenotfound-could-not-find-application-css
Heroku describe this as a slug cache issue. Clearing the cache using purge_cache can solve it.
Please have a look at this: https://github.com/heroku/heroku-repo#purge_cache

prathamesh-sonpatki:
Thanks @anuj-verma That was it! The code is deployed now

PR #642, opened 2017-11-14
remove Repo#path in favor of Repo#full_name, closes #587

### LINKED ISSUE #587: Why is `Repo#path` duplicating `Repo#full_name`?
[`Repo#path`](https://github.com/codetriage/codetriage/blob/b57ad8fac2fce0e920e3376693b106e8931de76d/app/models/repo.rb#L156) and [`Repo#full_name`](https://github.com/codetriage/codetriage/blob/b57ad8fac2fce0e920e3376693b106e8931de76d/app/models/repo.rb#L229) are identical, at least after the latter is set by a `before_save` callback. Why do these two exist? It was a bit confusing to me that there would be two ways of calling/finding very similar info, and it's not really clear to me why to have `full_name` given that the info for `path` is stored and can be easily referenced.

lostapathy:
This resolves the extra code noted on #587.  As noted there, full_name is not available until the record is persisted, but (at least now) the tests all pass and the app seems to work anyway so I went with it as-is.  If it turns out that's a problem, please point me in the right direction so I can add a test and adjust this to get it passing.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -32,7 +32,7 @@ class UserMailer < ActionMailer::Base
     return unless set_and_check_user(user)
     @repo  = repo
     @issue = assignment.issue
-    mail(to: @user.email, reply_to: "noreply@codetriage.com", subject: "Help Triage #{@repo.path} on GitHub")
+    mail(to: @user.email, reply_to: "noreply@codetriage.com", subject: "Help Triage #{@repo.full_name} on GitHub")
   end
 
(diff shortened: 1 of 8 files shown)


lostapathy:
Thanks - didn't think to look at the mailers.  I will update and try to write a test to make sure the mailers at least pass a basic smell test.

schneems:
Ready for review?

lostapathy:
Yes - sorry, I think this is ready now.

schneems:
Thanks!

PR #644, opened 2017-11-17
Clean pagination parameters that are not integers, resolves #581

### LINKED ISSUE #581: Deal with page parameters which are not integers
https://sentry.io/share/issue/3137333132372e333033363237373536/

lostapathy:
I elected to set bad values to nil rather than a value so that bad values fall through to
defaults and we don't have to keep them in sync.


lostapathy:
Closing this to clean up my list.  Feel free to re-open and merge in the future.

PR #648, opened 2017-11-22
Fewer SQL queries when assigning issues

schneems:
Previously if someone subscribed to get 3 issues from rails/rails then we would run 3 SQL queries. These queries are expensive. 

Instead, we can run 1 query and get 3 results and iterate through each result separately.

Other improvements:
- Use named binds in the query. Active Record has to convert the other query to use bind params, which takes time. I did benchmark/ips on the two methods and this one “won” but was not statistically signifiant.
- Freeze the sql string, so we don’t have to allocate a new one on every iteration.

## Apendix

I tried using `TABLESAMPLE` to decrease the overhead of the `random()`. While it helps on tables with lots of records. It is significantly slower for entries with fewer records.

### Lots of records (~5,000)

```
EXPLAIN
ANALYZE
SELECT
  *
FROM
  issues
WHERE
  repo_id = '2151' and
  state   = 'open' 
ORDER BY
  RANDOM()
LIMIT 1;

                                                                            QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=6187.56..6187.56 rows=1 width=1240) (actual time=8.812..8.812 rows=1 loops=1)
   ->  Sort  (cost=6187.56..6189.35 rows=3581 width=1240) (actual time=8.811..8.811 rows=1 loops=1)
         Sort Key: (random())
         Sort Method: top-N heapsort  Memory: 25kB
         ->  Bitmap Heap Scan on issues  (cost=97.63..6183.98 rows=3581 width=1240) (actual time=1.240..6.818 rows=5498 loops=1)
               Recheck Cond: ((repo_id = 2151) AND ((state)::text = 'open'::text))
               Heap Blocks: exact=3387
               ->  Bitmap Index Scan on index_issues_on_repo_id_and_created_at  (cost=0.00..97.46 rows=3581 width=0) (actual time=0.736..0.736 rows=5498 loops=1)
                     Index Cond: (repo_id = 2151)
 Planning time: 0.157 ms
 Execution time: 8.854 ms
(11 rows)


EXPLAIN
ANALYZE
SELECT
 * 
FROM 
  issues
WHERE
  id in (
  SELECT                                        
    id                                           
  FROM                                          
    issues 
  TABLESAMPLE SYSTEM(1)
  WHERE                                         
    repo_id = '2151' and                             
    state   = 'open' 
)
LIMIT 1;


                                                            QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.09..36.67 rows=1 width=1232) (actual time=1.810..1.810 rows=1 loops=1)
   ->  Nested Loop  (cost=0.09..1317.16 rows=36 width=1232) (actual time=1.809..1.809 rows=1 loops=1)
         ->  Sample Scan on issues issues_1  (cost=0.00..1169.82 rows=36 width=4) (actual time=1.797..1.797 rows=1 loops=1)
               Sampling: system ('1'::real)
               Filter: ((repo_id = 2151) AND ((state)::text = 'open'::text))
               Rows Removed by Filter: 5414
         ->  Index Scan using issues_pkey on issues  (cost=0.09..4.09 rows=1 width=1232) (actual time=0.009..0.009 rows=1 loops=1)
               Index Cond: (id = issues_1.id)
 Planning time: 0.292 ms
 Execution time: 1.847 ms
(10 rows)
```

TABLESAMPLE is faster

### Few records (~28 records)

```

EXPLAIN
ANALYZE
SELECT
  *
FROM
  issues
WHERE
  repo_id = '433' and
  state   = 'open' 
ORDER BY
  RANDOM()
LIMIT 1;
                                                                            QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=124.11..124.11 rows=1 width=1240) (actual time=0.088..0.088 rows=1 loops=1)
   ->  Sort  (cost=124.11..124.14 rows=68 width=1240) (actual time=0.087..0.087 rows=1 loops=1)
         Sort Key: (random())
         Sort Method: top-N heapsort  Memory: 25kB
         ->  Index Scan using index_issues_on_repo_id_and_created_at on issues  (cost=0.08..124.04 rows=68 width=1240) (actual time=0.019..0.072 rows=28 loops=1)
               Index Cond: (repo_id = 433)
 Planning time: 0.149 ms
 Execution time: 0.118 ms
(8 rows)

EXPLAIN
ANALYZE
SELECT
 * 
FROM 
  issues
WHERE
  id in (
  SELECT                                        
    id                                           
  FROM                                          
    issues 
  TABLESAMPLE SYSTEM(1)
  WHERE                                         
    repo_id = '433' and                             
    state   = 'open' 
)
LIMIT 1;
                                                         QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.09..1173.91 rows=1 width=1232) (actual time=24.947..24.947 rows=0 loops=1)
   ->  Nested Loop  (cost=0.09..1173.91 rows=1 width=1232) (actual time=24.946..24.946 rows=0 loops=1)
         ->  Sample Scan on issues issues_1  (cost=0.00..1169.82 rows=1 width=4) (actual time=24.945..24.945 rows=0 loops=1)
               Sampling: system ('1'::real)
               Filter: ((repo_id = 433) AND ((state)::text = 'open'::text))
               Rows Removed by Filter: 16267
         ->  Index Scan using issues_pkey on issues  (cost=0.09..4.09 rows=1 width=1232) (never executed)
               Index Cond: (id = issues_1.id)
 Planning time: 0.239 ms
 Execution time: 24.981 ms
(10 rows)
```

TABLESAMPLE is WAYYYY slower.

### DIFF
--- a/app/models/issue_assigner.rb
+++ b/app/models/issue_assigner.rb
@@ -9,46 +9,51 @@ class IssueAssigner
 
   def assign!
     subscriptions.each do |sub|
-      Array.new(sub.email_limit) do
-        assign_issue_for_sub(sub)
-      end
+      assign_issues_for_sub(sub, email_limit: sub.email_limit)
     end
     self
   end
 
   private
 
-  def assign_issue_for_sub(sub)
-    issue = Issue.find_by_sql(<<-SQL
+  def assign_issues_for_sub(sub, email_limit:)
+    issues = Issue.find_by_sql([%Q{
                 SELECT
                   *
                 FROM
                   issues
                 WHERE
-                  repo_id = '#{sub.repo_id}' and
-                  state   = '#{Issue::OPEN}'
+                  repo_id = :repo_id AND
+                  state   = :repo_state
                   AND id NOT IN (
                     SELECT
                       issue_id
                     FROM
                       issue_assignments
                     WHERE
-                      repo_subscription_id = '#{sub.id}'
+                      repo_subscription_id = :subscription_id
                   )
                 ORDER BY
                   random()
                 LIMIT
-                  1
-                SQL
-                             ).first
+                  :email_limit
+                }.freeze, {
+                  repo_id:         sub.repo_id,
+                  repo_state:      Issue::OPEN,
+                  subscription_id: sub.id,
+                  email_limit:     email_limit
+                }]).first(email_limit)
 
-    return false if issue.blank?
-    if issue.valid_for_user?(user)
-      sub.issue_assignments.create(issue_id: issue.id)
-    else
-      # prevent selecting this issue again and try to find another one
-      sub.issue_assignments.create(issue_id: issue.id, delivered: true) &&
-        assign_issue_for_sub(sub) # yay recursion!
+    return false if issues.blank?
+
+    issues.each do |issue|
+      if issue.valid_for_user?(user)
+        sub.issue_assignments.create(issue_id: issue.id)
+      else
+        # prevent selecting this issue again and try to find another one
+        sub.issue_assignments.create(issue_id: issue.id, delivered: true) &&
+          assign_issue_for_sub(sub, email_limit: 1) # yay recursion!
+      end
     end
   end
 end


schneems:
@prathamesh-sonpatki does the logic in this code look right to you?

lostapathy:
Logic looks good.

How often does the recursive call to get another record run in practice?  It might make sense to get 2*sub.email_limit (or some other factor) of random issues on the first query and just process the first sub.email_limit that are valid and discard the rest, to save trips to the db.  Would probably need to benchmark that against closer to production data than I have in my local env to tell if its a good tradeoff and how to tune it.

app/models/issue_assigner.rb
```diff
@@ -9,46 +9,51 @@ def initialize(user, subscriptions)
...
+    issues = Issue.find_by_sql([%Q{
                 SELECT
                   *
                 FROM
                   issues
                 WHERE
-                  repo_id = '#{sub.repo_id}' and
-                  state   = '#{Issue::OPEN}'
+                  repo_id = :repo_id AND
+                  state   = :repo_state
```
prathamesh-sonpatki: @schneems I think we should rename this to :issue_state, instead of :repo_state.

app/models/issue_assigner.rb
```diff
@@ -9,46 +9,51 @@ def initialize(user, subscriptions)
...
-                  1
-                SQL
-                             ).first
+                  :email_limit
+                }.freeze, {
+                  repo_id:         sub.repo_id,
+                  repo_state:      Issue::OPEN,
+                  subscription_id: sub.id,
+                  email_limit:     email_limit
+                }]).first(email_limit)
```
prathamesh-sonpatki: Is `first(email_limit)` required? It already returns `email_limit` rows right?

app/models/issue_assigner.rb
```diff
@@ -9,46 +9,51 @@ def initialize(user, subscriptions)
...
-    else
-      # prevent selecting this issue again and try to find another one
-      sub.issue_assignments.create(issue_id: issue.id, delivered: true) &&
-        assign_issue_for_sub(sub) # yay recursion!
+    issues.each do |issue|
+      if issue.valid_for_user?(user)
+        sub.issue_assignments.create(issue_id: issue.id)
+      else
+        # prevent selecting this issue again and try to find another one
+        sub.issue_assignments.create(issue_id: issue.id, delivered: true) &&
```
prathamesh-sonpatki: Should it be find_or_create, the issue_assignment might already exist.

prathamesh-sonpatki:
@schneems Logic looks good, added few comments.

prathamesh-sonpatki:
To avoid recursion, we can prolly include the joins with issue_assignments and check for such assignment is not present for the issues being selected. That will make the query complicated though.

PR #660, opened 2017-11-29
Schneems/cdn all the things

schneems:


schneems:
This doesn't work, because when a user is logged in, their status will never be seen by the server, that means for the first time an un-logged in user hits the homepage then the homepage will be cached and served by cloudfront. 

We would need to do something way more complex. For example setting up a subdomain like `active.codetriage.com` and when you log in you're redirected to that subdomain which would not be cached. But then you run into the issue of what happens when someone manually types in www.codetriage.com. In that case you would have to have some kind of JS that looks at cookies and dynamically redirects to `active.codetriage.com` (or something similar).

mattzollinhofer:
Ugh, I hate to be the guy who claims "flakey test" on his first PR, but the tests all pass locally. I can't see how my change would have affected the `user_mailer_test.rb`. Also, this previous build failure seems to be the same: https://travis-ci.org/codetriage/codetriage/jobs/409601461.

Thoughts?

schneems:
Was that comment for this PR? This is my old PR and should not be merged.

PR #661, opened 2017-12-02
Updated application port to 3000

budhrg:
Seems puma is running application in default port `3000`.

https://github.com/codetriage/codetriage/blob/f020070f0c44c79a90ff4f2c7c0e8ccb956a7f70/config/puma.rb#L14


schneems:
Thanks!

PR #665, opened 2017-12-21
update yard per github warning

lostapathy:
Github sent me a warning about an out of date version of yard.  PR'ing an update to satisfy that.


schneems:
Thanks. I’m on paternity leave but this was pretty simple.

PR #676, opened 2018-01-15
better docs for ApplicationController#set_title

johnfking:


bruno-b-martins:
Why not changing the method name to `set_web_page_title` or `set_page_head_title`?
Something self-explanatory that wouldn't need a comment to be understood.

PR #677, opened 2018-01-18
Filter by labels

hmasila:
This is a PR with the attempt to fix the issue stated here https://github.com/codetriage/codetriage/issues/498

- The first part(complete) is to filter repos by labels
- Lastly, filter issues by labels

image

image

PS//: Comments, reviews, and thoughts are welcome


app/assets/javascripts/homepage.js
```diff
@@ -21,6 +22,42 @@ $(document).ready(function(){
...
+    }).done(function(data) {
+      $('.repo-list-with-pagination').html(data["html"]);
+    }).fail(function(data){
+      console.log(data);
+    });
+
+    return false;
+  });
+})
+
```
hmasila: Yeah, I know this is duplicate code. Refactoring soon

kevincox:
It looks good to me. I didn't look too much into code style though. Some thoughts I had.
1. You tie labels to a repo. I wonder if people commonly want to listen to the same labels across repos, especially across an organization. 
1. Labels are identified by the name, which can change. I don't know if the github API supports tracking labels across renames but that would be cool if possible (although this would make point 1 infeasible) 

On a less important performance note. I'm not so sure about your indexes. I think most queries would filter by a specific label or two so you would probably want to search label -> issue. Rather then scan issues and check if the label matches.

(sidenote, the code sorting makes the diff quite hard to read, it would be nice to do it in a separate PR)

hmasila:
Hello, @kevincox Thanks for the review. I will look into the comments you have raised. However, I am not sure I follow what you mean by code sorting

kevincox:
I just meant that it looks like some for the code for database definitions was reordered. This puts a lot of deletions and (almost) identical additions which add a lot of noise to the diff.

hmasila:
Yeah sure. I will check this ones out

mario-d-s:
@hmasila this looks useful, so it would be a pity to have this go stale. Have you found time to address @kevincox's remarks?

hmasila:
@kevincox Regarding 

    You tie labels to a repo. I wonder if people commonly want to listen to the same labels across repos, especially across an organization.

I am not sure I understand what you mean here. Did you mean that the labels should be tied to a single repo and not all repos with issues that contain such a label?
If that is the case, then the filtering should not be done, say globally. Rather, it should be done when a user selects a Repo. That is i the `/repo` endpoint

hmasila:
In short move the filter from the root url and place it inside the https://github.com/codetriage/codetriage/blob/master/config/routes.rb#L58 endpoint

cc @mario-d-s

kevincox:
What I meant by that if you have repoa and repob which both have a label "foo" they will each have a different entry in the label table. I'm curious why not just have one entry for both repos, I don't see the benifit of diferentiating them.

However that is a small observation, I would be more concerned about label text changes.

hmasila:
Okay, I get you @kevincox I will also look into the label text change and update the PR accordingly

PR #690, opened 2018-03-06
[#close 658] Combined issues and docs + Testing Mail Question

schneems:
This PR does two fairly major things. First it incorporates issues and docs into the same email. The layout had to change fairly substantially for this to happen. All issues are now grouped by project.

The second major thing this PR does is add a layout to HTML emails. They look substantially better now :)

One big issue here is that even though sending this email is the single most important part of this app, it's utterly not tested. Many of the pieces are tested, but I've got very little confidence that the integration is properly done here. I can manually test via visiting http://localhost:3000/mail_view and going through various views, but I would love to have better tests here.

Any ideas?


chrisarcand:
⚠️ I've never used ActionMailer, just looking at things for fun ⚠️ 

At first glance I was going to say I'd try and test everything through the end of the responsibility boundary of the application, which is the call to `mail` from ActionMailer,  in a pseudo-integration sort of style and call it good -  but I note that apparently initializing mailer objects (which would be required for nice testing unless you want to resort to hacky things) is frowned upon and not idiomatic ActionMailer with this ['write instance methods and access them on the class level' API](https://github.com/rails/rails/blob/921b877c2fd51cc0abdcf0ef6ca3528808047056/actionmailer/lib/action_mailer/base.rb#L579-L581). Sigh.

It looks like ActionMailer does provide some nice things in `ActionMailer::TestCase` which you can use (including a `deliveries` array which contains the emails submitted for delivery in the test environment). So to me it makes sense to test your mailer sends the email you you expect it to via these helpers, and you can decide whether or not you want to do the full integration from the top with the job or if you want to test the components separately (I'd probably do the latter, personally).

test/emails/send_daily_triage_email_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'test_helper'
+
+class UserMailerTest < ActionMailer::TestCase
```
chrisarcand: Oooh TIL `ActiveJob::TestHelper::TestQueueAdapter`; these look great! 👍

PR #691, opened 2018-03-12
The “last” subscription may not have any issues

schneems:
This should be more representative of actual activity.


prathamesh-sonpatki:
@schneems Looks good to :shipit:

PR #693, opened 2018-03-19
Don't show repos in "Repos that need help" section when user is subscribed

thetizzo:
Hi,

I thought it would be useful to hide repos that a user is subscribed to from the "projects that need help" section so that they don't show up more than once on the user's dashboard when they are logged in.

### Current Behavior

As a user who is subscribed to Repo X, when I am logged in to the home page I see Repo X in the "Repos you are currently helping" section as well as in the "Open Source projects that need your help" section.

### Expected Behavior

As a user who is subscribed to Repo X,  when I am logged in I only see Repo X in the "Repos you are currently helping" section.

This PR is a working example.  Let me know if you think this is useful and I can finish it up and add tests.  I'm also open to any other feedback on implementation, code style, etc.

Thanks!

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -164,8 +164,8 @@ class Repo < ActiveRecord::Base
   end
 
   def self.without_user_subscriptions(user_id)
-    user_subscribed_repos = RepoSubscription.where(user_id: user_id).select(:repo_id)
-    self.where('id not in (?)', user_subscribed_repos)
+    user_subscribed_repo_ids = RepoSubscription.where(user_id: user_id).select(:repo_id)
+    self.where.not(id: user_subscribed_repo_ids)
   end
 
   def github_url
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -163,6 +163,11 @@ class Repo < ActiveRecord::Base
     self.where("issues_count > 0")
   end
 
+  def self.without_user_subscriptions(user_id)
+    user_subscribed_repos = RepoSubscription.where(user_id: user_id).select(:repo_id)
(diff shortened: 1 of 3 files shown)


app/models/repo.rb
```diff
@@ -163,6 +163,11 @@ def self.with_some_issues
     self.where("issues_count > 0")
   end
 
+  def self.without_user_subscriptions(user_id)
+    user_subscribed_repos = RepoSubscription.where(user_id: user_id).select(:repo_id)
+    self.where('id not in (?)', user_subscribed_repos)
```
prathamesh-sonpatki: can you use 

``` ruby
where.not(id: user_subscribed_repo_ids)
```

schneems:
Looks good. I approve after the change is made. Thanks for the PR and thanks for the review!

In the future i'm hoping to separate out these functionalities so that there is a dedicated page that shows all the repos you're subscribed to and allows you to do things like edit your subscriptions all on one page. Then I want to introduce a more powerful search page that allows you to do more than filter by language. For example sorting by subscriber count (desc or asc) etc.

thetizzo:
Thanks for the feedback!  I'm going to incorporate the changes and add tests this weekend.

thetizzo:
I've added a test for the `without_user_subscriptions` method that I added to the `Repo` class.  

Seems like this could use a system or integration test for the changes in `PagesController` but I'm not sure what to add for that.  I tried adding a test in `test/integrations/pages_controller.rb` to test the contents of `@repos` and `@repos_subs` for a signed in user but couldn't get anything to work reliably.  I can more tests for this but would appreciate some direction as to how.

Thanks!

schneems:
Thanks!

PR #696, opened 2018-03-27
Remove old skylight config

chancancode:
We are setting the SKYLIGHT_AUTHENTICATION ENV variable, so this is
not needed anymore. (It was from 2013!! 🎉 ❤️)

While I was at that, I also upgraded the skylight gem.


prathamesh-sonpatki:
💯

PR #702, opened 2018-04-10
Cache the description meta tag

vaidehijoshi:
Before this change, we were doing two queries per request to get the count of users and repos to generate the description meta tag (see screenshot below). Presumably, we don't actually care that these numbers are super up-to-date, so we can cache it up to an hour (same as the similar text in the [template's h2 tag](https://github.com/codetriage/codetriage/blob/633948dad71f072712dc322c514142f7df2f2a25/app/views/pages/index.html.slim#L3)).

[screen shot 2018-04-10 at 4 13 20 pm](https://oss.skylight.io/app/applications/IfSk4kDAh56r/1523380440/6h/endpoints/PagesController%23index?responseType=html)

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -16,7 +16,7 @@ class PagesController < ApplicationController
 
   def index
     set_title("Get Started Contributing to Open Source Projects")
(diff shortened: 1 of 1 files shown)


schneems:
Great job, thanks!

PR #715, opened 2018-06-12
Clarify in README that PostgreSQL is required

noahgibbs:
Clarify that Postgres is required, not just recommended. For instance, CodeTriage has Postgres-specific migrations like 20151214120117 (EnablePgStatExtension).


schneems:
Good call!

PR #725, opened 2018-07-27
Disallow Raw SQL by using Arel.sql for the raw part of the query

prathamesh-sonpatki:
- Followup of https://github.com/rails/rails/pull/27947
- Silences bunch of deprecation warnings.


schneems:
Thanks!

PR #726, opened 2018-07-27
Use simpler way to detect negative value

prathamesh-sonpatki:


schneems:
If we do this in a loop it will be expensive due to the array allocation.

PR #732, opened 2018-08-08
Add simple noscript message warning to enable JS

mattzollinhofer:
This is a quick stab at a `<noscript>` message. I wasn't sure where you'd want the the tag to live, `_nav` seemed like a reasonable spot, but let me know if you'd prefer something else.

 Also I didn't know if you'd object to me essentially hijacking the flash message styling for this warning. It seems to fit well, but I'd understand if you wanted to do that styling differently.

How it should look:
noscript-screenshot

A link to easily disable javascript in chrome dev tools in case it's not top of mind: https://stackoverflow.com/questions/13405383/how-to-disable-javascript-in-chrome-developer-tools

### DIFF
--- a/app/views/application/_nav.html.slim
+++ b/app/views/application/_nav.html.slim
@@ -1,3 +1,6 @@
+noscript
+  .alert.flash-error CodeTriage requires JavaScript, please enable it!
+
 .top-bar-red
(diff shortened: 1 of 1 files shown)


schneems:
Thanks!

PR #734, opened 2018-08-09
Address deprecation warning

mattzollinhofer:
The warning says:

>   DEPRECATION WARNING: Dangerous query method (method whose arguments
>   are used as raw SQL) called with non-attribute argument(s): "random()".
>   Non-attribute arguments will be disallowed in Rails 6.0. This method
>   should not be called with user-provided values, such as request
>   parameters or model attributes. Known-safe values can be passed by
>   wrapping them in Arel.sql().

There's currently only one deprecation warning in the test suite, but it
seems like a fairly straightforward change to make throughout as there
was no user input being used, just plain RANDOM() calls.


schneems:
Thanks again!

PR #736, opened 2018-08-10
Sort the select options after converting to local time

### LINKED ISSUE #721: "Email time of day" account option ordering is incorrect.
Under account options, there is a box which allows you to select when the email is sent during the day. The ordering for these times could use some help.

Currently it looks like:

```
19:00 EST
20:00 EST
21:00 EST
22:00 EST
23:00 EST
00:00 EST
01:00 EST
02:00 EST
03:00 EST
04:00 EST
05:00 EST
06:00 EST
07:00 EST
08:00 EST
09:00 EST
10:00 EST
11:00 EST
12:00 EST
13:00 EST
14:00 EST
15:00 EST
16:00 EST
17:00 EST
18:00 EST
```

It looks like 19-23 are being placed out of order for some reason.

mattzollinhofer:
The current implentation leaves the original order (UTC) but just
changes the labels so it appears out of order to the user. The
proposed change is not amazing code, but seem effective to order based
on the user's local time.

Advantages: quick, easy, effective
Disadvantages: not the prettiest code, breaks handlers I believe (but
there are none that I see)

I thought this might be a good time to pick a time as the default. It
seems unlikely that midnight UTC is many people's first choice, so I
figured after work-ish (7PM local) would be a reasonable default.

References:
 - jquery select sort: https://stackoverflow.com/a/7466196

Resolves #721.

### DIFF
--- a/app/assets/javascripts/user_email_time_of_day.js
+++ b/app/assets/javascripts/user_email_time_of_day.js
@@ -9,4 +9,11 @@ $(document).ready(function(){
       element.text = LocalTime.strftime(time, "%H:%M %Z");
     })
 
+  $("#user_email_time_of_day").html($("#user_email_time_of_day option").sort(function (a, b) {
+    return a.text == b.text ? 0 : a.text < b.text ? -1 : 1
+  }))
+
+  var after_work = $("#user_email_time_of_day").find("option:contains('19:00')").val()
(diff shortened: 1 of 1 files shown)


schneems:
Thanks!

schneems:
You’re killing it with these PRs!

PR #737, opened 2018-08-11
Update formatting for branch names

### LINKED ISSUE #688: Git branch name is invalid
I just worked on a documentation fix where it suggested my branch name have a `?` in it. This character isn't valid for git branches. The suggested branch name should probably filter out invalid characters. 

Other than that, I truly love this site!

mattzollinhofer:
Resolves #688 which reports an issue when there is a question mark in the
method name. git spec states that question marks are not allowed. I took
more conservative approach and only allowed numbers, letters, and
underscores.

I also converted periods to pounds:
  `Payment.charge` -> `Payment#charge`

### DIFF
--- a/app/controllers/doc_methods_controller.rb
+++ b/app/controllers/doc_methods_controller.rb
@@ -6,10 +6,8 @@ class DocMethodsController < ApplicationController
                         .first
     @comment = @doc.doc_comments.select(:comment).first
     @repo    = @doc.repo
-
-    # http://stackoverflow.com/questions/3651860/which-characters-are-illegal-within-a-branch-name
     @username = current_user.present? ? current_user.github : "<your name>"
-    @branch   = "#{@username}/update-docs-#{@doc.path}-for-pr".gsub(/:|~|\^|\\|\.\./, "_")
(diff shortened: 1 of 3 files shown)


schneems:
Thanks a ton for this!

In the shell anything after the pound will be interpreted as a comment. I think we need another character or keeping the period seems fine.

Could you move the logic into a model or helper and then add a test?

mattzollinhofer:
Ha, good point about pound in shell, wasn’t thinking. I was only changing it because officially branch names also can’t have two periods in a row (‘git diff invalid..branchname..master’) and got cute with the pound:-). I’d lean towards just replacing periods like everything else and only allowing letters and numbers but I’d be happy to leave them and hope there no two in a row.  Preference?

Yeah I’ll pull that out and get some tests around it too.

mattzollinhofer:
I updated and removed periods, let me know if you'd prefer leaving them in.

I also decided to make a little poro rather than moving into a model or helper. I've been moving towards small objects like this in my dev work, but I know not everyone prefers that. So just say the word and I'll pull the method into a model or helper.

schneems:
Awesome! Thanks!

PR #738, opened 2018-08-12
Revert default email time of day selection

mattzollinhofer:
* I introduced a bug in the way that I selected the email time of day
because I was ALWAYS selecting the default time instead of only when the
user hadn't selected a date.

* There are a number of ways to do this, but it's not something anyone
asked for (or that anyone probably needs). No code being better than
more code let's just remove it for now?


schneems:
Seems good

Issue #681, opened 2018-01-31
Add subscribers count to repos#show subscribers fragment cache


schneems:
After removing yourself from a repo, the repo will continue to show you as a subscriber due to the cache. We could fix this by having the key depend on the subscribers count.

PR #739, opened 2018-08-12
Force repo show page to invalidate cache

mattzollinhofer:
I'll start this by saying that caching in rails is NOT a strength of mine. I figured solving this issue would give me the chance to get a bit smarter on the matter. Fortunately/unfortunately I got it to work sooner than I thought.

Fortunately: it seems to work.
Unfortunately: I feel like it was too easy so I'm afraid it's not correct 😄.

As you see, all I did was add `touch: true` in `RepoSubscription` on the `belongs_to: repo` association.

The obvious thing I could have messed up was just testing it incorrectly in dev mode, but I did set `rails dev:cache` (thanks rails, that's quite nice). So I think I got that part right. I manually tested this by subscribing/unsubscribing after setting the cache mode in dev with and without the `touch: true` beings set so I _think_ it works, but I lack full confidence in this area so take a second look at this one please! 😄 

I'm happy to keep reading on caching if this is not the right approach or if this has some side effect I'm not aware of...

### DIFF
--- a/app/models/repo_subscription.rb
+++ b/app/models/repo_subscription.rb
@@ -6,7 +6,7 @@ class RepoSubscription < ActiveRecord::Base
   validates  :user_id, presence: true
   validates  :email_limit, numericality: { less_than: 21, greater_than_or_equal_to: 0 }
 
-  belongs_to :repo, counter_cache: :subscribers_count
+  belongs_to :repo, counter_cache: :subscribers_count, touch: true
   belongs_to :user
 
   has_many   :issue_assignments


schneems:
Great idea!

mattzollinhofer:
I don't think you looked closely enough at this 😆 

I meant to also say that this is response to your issue #681, so that can close if this works as desired.

schneems:
I’ve actually not seen the touch option until today. Looks like it solves
the immediate issue.

My only concern is that this change might cause extra stress on the
database. Only time and “heroku pg:outliers” will tell. I don’t think it
will be bad though.

schneems:
Ahh, i think that may a different issue. You can set yourself to a "private" account so we don't show your face in the subscriber list, but we still count you for the overall value in the badge.

mattzollinhofer:
Gotcha, I could see that for the first part, but the number should update when I subscribe and then unsubscribe which I’m not right now.

PR #741, opened 2018-08-14
accessibility improvement: Add text to logo

mattzollinhofer:
Visually you shouldn't see a difference, but a screen reader will now tell the user that the link is to "Code Triage Home". I think that will be an improvement.

 * Style taken from bourbon version 5.1.0 to allow for providing text to screen readers but hiding for visual users. This should be removed when bourbon is upgraded.
   https://github.com/thoughtbot/bourbon/blob/v5.1.0/core/bourbon/library/_hide-visually.scss


schneems:
Thanks!

PR #742, opened 2018-08-14
accessibility enhancement: hierarchical heading tags

mattzollinhofer:
Update heading tags to be hierarchical. `h4` -> `h3`. The styling looks the same to me visually and now the page structure will not skip the h3 tags.


schneems:
thanks!

PR #763, opened 2018-08-27
Bump bourbon from 4.2.5 to 5.1.0 (replacement for #762)

greysteil:
Replacement for #762 which handles the breaking changes in Bourbon 5.

Needs testing carefully! I've followed the upgrade guide for Bourbon 5, but haven't run locally.


schneems:
Awesome! Thanks! I'll look  at this soon. One question. When does depbot decide to upgrade dependencies? For example I recently found out there were newer versions of dalli and pg and upgraded them manually. Any idea why dependabot would have skipped those ?

schneems:
Looks good locally. Thanks!

greysteil:
For this repo you've got the update schedule set to `monthly`, so it'll create PRs once a month, on the 1st. The exception is dependencies with a security vulnerability - it will create PRs for those immediately (and tag them as security issues).

PR #779, opened 2018-09-16
Call branchname on GitBranchnameGenerator to get branchname

shanecav84:
The `doc_methods#show` view is showing the `GitBranchnameGenerator` object itself, not the branch name, because `branchname` isn't being called on the object. This PR adds a call to `branchname`.


schneems:
whoops. thanks!

mattzollinhofer:
Ugh that’s actually my bad. TFW you do a nice little refactor and then don’t actually call the method :-/

schneems:
My personal favorite is when I forget to call `new` and instead try passing arguments directly to a constant i.e. `GitBranchnameGenerator(username: @username, doc_path: @doc.path).branchname`

PR #802, opened 2018-10-07
Poking inactive users based on favorite language(s)

### LINKED ISSUE #701: Emailed issues don't use language prefs
In email that I received, containing list of issues, the project is in a programming language (c#) that I did not checkmark in my profile. I think this should be a bug rather than a feature. As it stands it is too likely that the issue received will not be of interest since there are 50 Lang's but only 5'ish that interest me so basically 90% chance of irrelevant issues mentioned in email.

romanrizzi:
## Description

As a user, I want to receive repos where I can contribute to. 

## Related Issue
Closes #701

## Motivation and Context

When poking inactive users, we should consider their favorite languages so there're better chances of engagement.

## How Has This Been Tested?

- Missing tests poke inactive users job were added.
- Tested that repos were taking favorite language(s) into account with some unit tests.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/app/jobs/background_inactive_email_job.rb
+++ b/app/jobs/background_inactive_email_job.rb
@@ -4,13 +4,19 @@ class BackgroundInactiveEmailJob < ApplicationJob
   def perform(user)
     return false if user.repo_subscriptions.present?
 
-    most_issues_repo = Repo.order_by_issue_count.first
-    repo_in_need = Repo.order_by_need.not_in(most_issues_repo.id).first
+    with_most_issues = curated_repos_for(user).order_by_issue_count.first
+    in_need = curated_repos_for(user).order_by_need.not_in(with_most_issues.id).first
     UserMailer.poke_inactive(
       user: user,
-      most_issues_repo: most_issues_repo,
-      repo_in_need: repo_in_need,
-      random_repo: Repo.rand.not_in(most_issues_repo.id, repo_in_need.id).first
+      most_issues_repo: with_most_issues,
+      repo_in_need: in_need,
+      random_repo: curated_repos_for(user).rand.not_in(with_most_issues.id, in_need.id).first
     ).deliver_later
+  end
+
+  private
+
+  def curated_repos_for(user)
+    Repo.select(:id, :full_name).in_user_lang_preferences(user)
   end
 end
--- a/app/jobs/background_inactive_email_job.rb
+++ b/app/jobs/background_inactive_email_job.rb
@@ -6,11 +6,13 @@ class BackgroundInactiveEmailJob < ApplicationJob
 
     with_most_issues = curated_repos_for(user).order_by_issue_count.first
     in_need = curated_repos_for(user).order_by_need.not_in(with_most_issues.id).first
+    random = curated_repos_for(user).rand.not_in(with_most_issues.id, in_need.id).first
+
     UserMailer.poke_inactive(
       user: user,
       most_issues_repo: with_most_issues,
       repo_in_need: in_need,
-      random_repo: curated_repos_for(user).rand.not_in(with_most_issues.id, in_need.id).first
+      random_repo: random
     ).deliver_later
   end
 
--- a/app/jobs/background_inactive_email_job.rb
+++ b/app/jobs/background_inactive_email_job.rb
@@ -3,6 +3,14 @@
 class BackgroundInactiveEmailJob < ApplicationJob
   def perform(user)
     return false if user.repo_subscriptions.present?
-    UserMailer.poke_inactive(user: user).deliver_later
+
+    most_issues_repo = Repo.order_by_issue_count.first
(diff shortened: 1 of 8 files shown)


test/unit/repo_test.rb
```diff
@@ -126,4 +126,44 @@ class RepoTest < ActiveSupport::TestCase
...
     assert repos.include?(unsubscribed_repo)
   end
+
+  # test '.order_by_issue_count' do
+  #  expected_order_by_issues = [repos(:issue_triage_sandbox), repos(:rails_rails)]
+
+    # repos = Repo.order_by_issue_count
+
+    # assert_equal repos, expected_order_by_issues
+  # end
```
romanrizzi: Tried adding some tests for other related queries as well. I faced some complications because the database already have existing data. Would it be reasonable to add a gem like [DatabaseCleaner](https://github.com/DatabaseCleaner/database_cleaner) to perform a teardown between tests?

app/models/repo.rb
```diff
@@ -170,6 +170,11 @@ def self.without_user_subscriptions(user_id)
     self.where.not(id: user_subscribed_repo_ids)
   end
 
+  def self.in_user_lang_preferences(user)
+    return all unless user.has_favorite_languages?
+    self.where('language IN (?)', user.favorite_languages)
```
oniofchaos: Shouldn't need to go to raw SQL for this,

```ruby
self.where(language: user.favorite_languages)
```
should produce the same result and keep this in full ActiveRecord.

app/views/user_mailer/poke_inactive.html.erb
```diff
@@ -3,16 +3,22 @@ Hello <%= @user.github %>,<br /><br />
...
 
-Help a random repo:
-<h3><%= link_to @random_repo.full_name, repo_url(@random_repo) %></h3><br />
+<% if @random_repo.present? %>
+  Help a random repo:
+  <h3><%= link_to @random_repo.full_name, repo_url(@random_repo) %></h3><br />
+<% end %>
 
-If you don't like any of those <%= link_to "add your own", new_repo_url %> repo, <%= link_to "browse existing " , root_url %> repos, or you can even subscribe to <%= link_to "CodeTriage itself", File.join(root_url, "/codetriage/codetriage") %>. The important part is that you pick something to triage, or you will get this same email next week, maybe with different repos.<br /><br />
+If yoy don't like any of these options, remember that they are selected based in your language preferences (<%= link_to 'click here to change them', user_url(@user) %>). Not what you're looking for? <%= link_to "add your own", new_repo_url %> repo, <%= link_to "browse existing " , root_url %> repos, or you can even subscribe to <%= link_to "CodeTriage itself", File.join(root_url, "/codetriage/codetriage") %>. The important part is that you pick something to triage, or you will get this same email next week, maybe with different repos.<br /><br />
```
oniofchaos: `yoy` => `you`

`based in your` => `based on your`

thoughts?

app/jobs/background_inactive_email_job.rb
```diff
@@ -1,6 +1,20 @@
 class BackgroundInactiveEmailJob < ApplicationJob
   def perform(user)
     return false if user.repo_subscriptions.present?
-    UserMailer.poke_inactive(user: user).deliver_later
+
+    with_most_issues = curated_repos_for(user).order_by_issue_count.first
```
oniofchaos: To me, overall this logic feels kinda complicated for a job. I'm not super familiar with this codebase, but something like a method on the `Repo` model or a `CuratedRepoService` would be how I would approach this - I like to keep my jobs pretty slim though.

Thoughts?

app/views/user_mailer/poke_inactive.html.erb
romanrizzi: Sounds cool.

romanrizzi:
Thanks for the review @oniofchaos!

I'll be taking care of these suggestions over the weekend 👍

app/jobs/background_inactive_email_job.rb
romanrizzi: The logic itself is already encapsulated inside the scopes and we're just chaining them.

`#curated_repos_for` is just the result of extracting duplicating code into a method to improve readability.

Not only replacing these calls by a service object call will add another layer of indirection but also it'll not be used by any other part of the code. I think we should wait to need the same behavior in a different place before extracting it.

romanrizzi:
It was a really long weekend 😜 

I addressed your comments @onionofchaos, thanks again!

PR #811, opened 2018-10-10
Default avatar for private users on repo show page

### LINKED ISSUE #683: Default avatar for private users on repo show page
A repo show page might show 1 subscriber, but no avatars if the only subscriber is "private". In this case either show a default URL and do not link to anything.

goronfreeman:
Resolves #683 

This adds a default avatar image for private users that are subscribed to a repo. I used [this](https://material.io/tools/icons/?icon=account_circle&style=baseline) icon from Google's material design iconset, because it's free to use (under [Apache license version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html)) and looks nice in my opinion. I _think_ this is what was being asked for in the issue description, but the wording was a little confusing.

I didn't add any tests, because I'm not sure if tests are normally added for this sort of thing. If anyone thinks there should be a test, I can add one.


oniofchaos:
@goronfreeman definitely 👍 to tests IMO

@schneems does this fix the issue in #683?

goronfreeman:
@oniofchaos Just noticed this PR as well. I just took a look at the existing tests, and I'm exactly sure where a test for this would fit in. There aren't any Capybara tests that I could find for checking that the default avatar is on the page, so not sure if I should be the one to add the first test. Let me know what you think.

oniofchaos:
I don't see a problem with adding a test if there isn't one - leave things better than you found them :)

schneems:
Thanks!

PR #812, opened 2018-10-12
Add target option in github links

makketagg:
Hey, everyone. Added `target` option for GitHub links.


mattzollinhofer:
Adding target _blank changes default browser behavior. I’ve always thought it best to leave browser behavior alone unless there’s a really good reason. If someone wants to open a new tab/window they can do so on their own. 

A blog that I found helpful about the topic: https://css-tricks.com/use-target_blank/

PR #866, opened 2018-12-10
Upgraded Ruby version in .travis.yml and Gemfile to 2.5.1

jhsu802701:
Gemfile.lock specifies Ruby 2.5.1, but Travis still specifies 2.5.0.  This pull request resolves this discrepancy by bumping Ruby to 2.5.1 in the Travis environment.  Additionally, this pull request pins the Ruby version in the Gemfile at 2.5.1.  In other words, this pull request makes the Ruby version specification (2.5.1) consistent across the Gemfile, Gemfile.lock, and .travis.yml files.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-ruby '>= 2.5', '< 2.7'
+ruby '2.5.1'
 
 git_source :github do |name|
   "https://github.com/#{name}.git"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,7 +5,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
-ruby '>= 2.5', '< 2.7'
+ruby '2.5.1'
```
prathamesh-sonpatki: I think @schneems added this so that we can experiment with Ruby trunk or Ruby 2.6 previews without changing Gemfile all the time.

jhsu802701:
I will submit a new pull request that updates .travis.yml only and leaves the Gemfile alone.

PR #867, opened 2018-12-10
Pinned Ruby version in Gemfile to 2.5.1

jhsu802701:
This makes it consistent with the Ruby version specified in Gemfile.lock.


jhsu802701:
I'm removing this pull request and adding the change in this pull request to my other one that consists of upgrading Ruby in Travis.  The change here will not work in Travis until .travis.yml specifies Ruby 2.5.1 (instead of 2.5.0).

PR #872, opened 2018-12-10
Upgraded Ruby to 2.5.1 in .travis.yml

jhsu802701:
Gemfile.lock specifies Ruby 2.5.1.  This pull request makes the .travis.yml configuration consistent with that specified by Gemfile.lock.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ branches:
   only: [master]
 
 rvm:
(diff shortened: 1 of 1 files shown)


jhsu802701:
Given that the Ruby cersion has since been upgraded to 2.6.0 for both Travis and the Gemfile, this pull request is no longer necessary.

PR #873, opened 2018-12-10
Allow SimpleCov to work in the local environment

jhsu802701:
Under the current test setup, SimpleCov does not work in the local development environment.  This pull request corrects this.


jhsu802701:
I corrected the merge conflict.

schneems:
Can you blame the original commit and see why it was wrapped in a conditional?

jhsu802701:
The activation of SimpleCov has always been wrapped in the conditional.  No explanation for this was ever provided.

https://github.com/codetriage/codetriage/commit/86288b286b50158bf1d2bacb109f5aee933881f8#diff-f14ca9959ff06f3ca2ee6b9564866954

schneems:
Thanks maybe it was related to test time? any idea if generating code coverage impacts test time?

What does running tests with the “time rake test” command generate with and without the env var?

Also, is it not good enough for your case to SE the env var when running locally?

jhsu802701:
Generating code coverage reports doesn't seem to affect test time.  I see no reason to omit this in the local development environment.  Additionally, the test coverage isn't done in the Travis CI environment either.

Additionally, there should be a code coverage badge.  (Since I'm not a collaborator, I cannot provide this.)

schneems:
Thanks!

PR #874, opened 2018-12-10
Added brakeman, bundler-audit, and gemsurance

jhsu802701:
Added brakeman, bundler-audit, and gemsurance

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,9 +72,12 @@ group :test do
 end
 
 group :development, :test do
+  gem 'brakeman', '4.3.1'
+  gem 'bundler-audit', '0.6.0'
   gem 'derailed_benchmarks'
   gem 'dotenv-rails'
   gem 'faker', require: false
+  gem 'gemsurance', '0.9.0'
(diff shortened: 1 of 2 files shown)


jhsu802701:
I find the brakeman and bundler-audit gems to be valuable for flagging possible security vulnerabilities.  I find gemsurance to be essential for flagging the gems that need to be updated because they're outdated or have security vulnerabilities.

strika:
These gems are indeed great. However, adding them to the Gemfile does have a drawback - a gem version can easily become outdated and in my experience, it takes a while for developers to update them. That means that the security advices that were added to the gem are not applied.

I found that it's best to make a simple script that install the latest brakeman/bundler-audit every time and run checks. Like this:

```bash
#!/usr/bin/env bash

gem install brakeman --no-ri --no-rdoc
brakeman -z --table-width 300
```

jhsu802701:
Given the team's desire to minimize the number of gems used, I'm adding these gems to my special personal branch but closing this pull request.

PR #893, opened 2019-01-12
Corrected subscriber_count to subscribers_count; added integration tests of user public/private status

jhsu802701:
I added integration tests of the user's ability to switch between public and private status and the user's visibility on a repo page.  (At the moment, private subscribers are not shown on the repo pages but public subscribers are.)

Many parts of the code call the subscribers_count method for the repo object, but it was called "subscriber_count" in the repo model and in a few other places.  I have replaced all instances of "subscriber_count" with "subscribers_count".  The tests of the visibility of a repo's subscribers were failing before I made this change.


app/models/repo.rb
```diff
@@ -106,7 +106,7 @@ def weight
     end
   end
 
-  def subscriber_count
+  def subscribers_count
```
oniofchaos: What do you think about making this a delegate instead of a method?

jhsu802701:
I resolved the merge conflict.

PR #895, opened 2019-01-13
Added basic repo model tests (weight and color)

jhsu802701:


test/unit/repo_test.rb
```diff
@@ -126,4 +126,53 @@ class RepoTest < ActiveSupport::TestCase
...
+    assert_equal @repo.weight, ''
+    assert_equal @repo.color, 'lightgrey'
+
+    @repo.issues_count = 0
+    assert_equal @repo.weight, 'low'
+    assert_equal @repo.color, '5bb878'
+
+    @repo.issues_count = 100
+    assert_equal @repo.weight, 'low'
+    assert_equal @repo.color, '5bb878'
```
avcwisesa, line 148: I think for this case, creating tests for the minimum & maximum value for each weight category threshold will represent the function clearly enough. I would suggest only keeping the test for:
-1, 0, 199, 200, 799, 800

PR #898, opened 2019-01-13
Added application helper test

jhsu802701:
Added tests of the warning_svg, star_svg, and flash_class helpers.

### DIFF
--- a/test/helpers/application_helper_test.rb
+++ b/test/helpers/application_helper_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+
+class ApplicationHelperTest < ActionView::TestCase
+  test 'warning_svg' do
+    assert_equal warning_svg, WARNING_SVG
+  end
+
+  test 'star_svg' do
+    assert_equal star_svg, STAR_SVG
+  end
+
+  test 'flash_class levels' do
+    assert_equal flash_class(:notice), 'info'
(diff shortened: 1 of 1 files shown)


test/helpers/application_helper_test.rb
```diff
@@ -0,0 +1,17 @@
+require 'test_helper'
+
+class ApplicationHelperTest < ActionView::TestCase
+  test 'warning_svg' do
```
prathamesh-sonpatki: This is just testing the method against the constant. It will pass even when the constant is modified. So we can skip this test and the test for star_svg.

test/helpers/application_helper_test.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  test 'star_svg' do
+    assert_equal star_svg, STAR_SVG
+  end
+
+  test 'flash_class levels' do
+    assert_equal flash_class(:notice), 'info'
+    assert_equal flash_class(:error), 'error'
+    assert_equal flash_class(:alert), 'warning'
```
prathamesh-sonpatki: The expected value should come first in the assertion.

``` ruby
assert_equal 'info', 'flash_class(:notice)'
```

jhsu802701:
I am replacing this pull request with #932 .

PR #902, opened 2019-01-13
Updated the user show capability and tests

jhsu802701:
Viewing a user's profile while not logged in or while logged in as a different user results in being redirected to the home page.  Controller and integration tests of this are added.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -3,9 +3,7 @@ class UsersController < ApplicationController
 
   def show
     @user = User.find(params[:id])
-    if @user.private?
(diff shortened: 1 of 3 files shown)


bruno-b-martins:
Hey there,
It seems like your solution doesn't allow for others to see other public users.
Weren't you trying to make that possible?
Cheers!

jhsu802701:
Until we decide on what user information should be available to other users, I think it's better to keep the profiles of public users private.

jhsu802701:
Under the status quo, a user who attempts to view another user's profile just sees a blank page, which sounds pointless to me.

PR #907, opened 2019-02-03
Added tests for issues and doc links

GabrielRMuller:
Simple tests created to analyze the basics functionalities that these links have to provide. The issue test checks if it contains a link and if the link leads to a github page. The doc test checks if the link will lead to the doc_methods page.

### DIFF
--- a/test/integration/maintaining_repo_subscriptions_test.rb
+++ b/test/integration/maintaining_repo_subscriptions_test.rb
@@ -34,6 +34,17 @@ class MaintainingRepoSubscriptionsTest < ActionDispatch::IntegrationTest
     assert page.has_content?("mockstar")
   end
 
+  test "clicking an issue leads to a github page" do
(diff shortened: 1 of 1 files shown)


test/integration/maintaining_repo_subscriptions_test.rb
```diff
@@ -34,6 +34,17 @@ def triage_the_sandbox
     assert page.has_content?("mockstar")
   end
 
+  test "clicking an issue leads to a github page" do
+    triage_the_sandbox
+    assert page.has_link?("first test issue", :href => /github.com/)
```
marclerodrigues: What do you think about using the Ruby 1.9 hash syntax here?
```ruby
href: /github.com/
```

GabrielRMuller: good idea, should've done that in the first place :smile:

Issue #801, opened 2018-10-05
Random repo in user mailer can sometimes be nil


nateberkopec:
This randomly fails on tests and in prod. 

[Problem is this line.](https://github.com/codetriage/codetriage/blob/master/app/mailers/user_mailer.rb#L78)

GabrielRMuller:
Hey! I was taking a look at this test and the reason it was failing was probably because sometimes there wasn't a repo that wasn't in both `@most_repo` and `@need_repo`, so using `not_in` to search for one was the cause of the `nil` in `@random_repo`. 

To fix this I removed `not_in`and now it searches for a random repo using only `rand` (also I fixed a small typo on the `repo.rb` file). Take a look! #909

prathamesh-sonpatki:
Closed by #909

PR #909, opened 2019-02-08
fix intermittent test in user_mailer_test

GabrielRMuller:
changed the way @random_repo gets a random user so it never returns a `nil` and consequently fixing the intermittent error on the test.

Also fixed a tiny typo in `repo.rb` :smile:  #801

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -75,7 +75,7 @@ class UserMailer < ActionMailer::Base
     return unless set_and_check_user(user)
     @most_repo   = Repo.order_by_issue_count.first
     @need_repo   = Repo.order_by_need.not_in(@most_repo.id).first
-    @random_repo = Repo.rand.not_in(@most_repo.id, @need_repo.id).first
+    @random_repo = Repo.rand.not_in(@most_repo.id, @need_repo.id).first || @most_repo || @need_repo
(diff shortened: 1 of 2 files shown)


app/mailers/user_mailer.rb
```diff
@@ -75,7 +75,7 @@ def poke_inactive(user:)
     return unless set_and_check_user(user)
     @most_repo   = Repo.order_by_issue_count.first
     @need_repo   = Repo.order_by_need.not_in(@most_repo.id).first
-    @random_repo = Repo.rand.not_in(@most_repo.id, @need_repo.id).first
+    @random_repo = Repo.rand.first
```
prathamesh-sonpatki: @GabrielRMuller The purpose here is not send the inactive user a repo which needs most support. So we can use a fallback here instead of selecting completely randomly. We can check if a random repo which is not needing support and only if it does not exist, we can send the most_repo or need_repo.

``` ruby
@random_repo = Repo.rand.not_in(@most_repo.id, @need_repo.id).first || @most_repo || @need_repo
```

GabrielRMuller: good call! this keeps the code logic more similar to the one intended originally. I've already committed the change.

prathamesh-sonpatki:
Thanks ❤️

Issue #862, opened 2018-11-27
Improve test coverage


jhsu802701:
SimpleCov shows only 68% test coverage.  The weakest points are libraries (19%) and controllers (44%).

Mehonoshin:
@jhsu802701 feel free to open PRs and add missing specs

jhsu802701:
I submitted a pull request (https://github.com/codetriage/codetriage/pull/873) to enable SimpleCov in the development environment.

I also believe that this app needs a test coverage badge.  Is there any specific reason that there isn't one?  (If so, I will add one to my fork.)

jhsu802701:
I've made several pull requests that fill in missing tests.

PR #930, opened 2019-03-22
Add test coverage for app/controllers/university_controller.rb

laka3000:
Related to #862

Including tests for app/controllers/university_controller.rb

### DIFF
--- a/test/functional/university_controller_test.rb
+++ b/test/functional/university_controller_test.rb
@@ -1,0 +1,50 @@
+require 'test_helper'
+
+class UniversityControllerTest < ActionController::TestCase
+  include Devise::Test::ControllerHelpers
+
+  test 'when picking a repo' do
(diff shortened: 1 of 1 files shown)


test/functional/university_controller_test.rb
```diff
@@ -0,0 +1,50 @@
+require 'test_helper'
+
+class UniversityControllerTest < ActionController::TestCase
+  include Devise::Test::ControllerHelpers
+
+  test 'when picking a repo' do
+    get :show, params: { id: 'picking_a_repo' }
+
+    assert_response :success, 'successfully renders show'
+    assert_not_nil assigns(:page_title), 'assigns page title'
```
prathamesh-sonpatki: Usage of `assigns` and `assert_template` is not encouraged Rails 5 onwards. So let's not use it.

PR #945, opened 2019-03-31
Improvise ActiveRecord queries

anmolarora:
- ActiveRecord query fixes

### DIFF
--- a/app/models/issue_assigner.rb
+++ b/app/models/issue_assigner.rb
@@ -21,28 +21,10 @@ class IssueAssigner
   private
 
   def assign_issue_for_sub(sub)
-    issue = Issue.find_by_sql(<<-SQL
-                SELECT
-                  *
-                FROM
-                  issues
-                WHERE
-                  repo_id = '#{sub.repo_id}' and
-                  state   = '#{Issue::OPEN}'
-                  AND id NOT IN (
-                    SELECT
-                      issue_id
-                    FROM
-                      issue_assignments
-                    WHERE
-                      repo_subscription_id = '#{sub.id}'
-                  )
-                ORDER BY
-                  random()
-                LIMIT
-                  1
-                SQL
-                             ).first
+    issue = Issue.where(repo_id: sub.repo_id, state: Issue::OPEN)
+                 .where.not(id: IssueAssignment.select(:issue_id).where(repo_subscription_id: sub.id))
+                 .order(Arel.sql('RANDOM()'))
+                 .first
 
     return false if issue.blank?
     if issue.valid_for_user?(user)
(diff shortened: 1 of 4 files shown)


anmolarora:
@prathamesh-sonpatki Thoughts?

app/models/issue_assigner.rb
```diff
@@ -19,28 +19,10 @@ def assign!
...
-                    WHERE
-                      repo_subscription_id = '#{sub.id}'
-                  )
-                ORDER BY
-                  random()
-                LIMIT
-                  1
-                SQL
-                             ).first
+    issue = Issue.where(repo_id: sub.repo_id, state: Issue::OPEN)
```
tiagofsilva: Since you're refactoring this query, the `Issue` model has a "scope" `.open_issues` to use here.

app/models/issue_assigner.rb
```diff
@@ -19,28 +19,10 @@ def assign!
...
+                 .where.not(id: IssueAssignment.select(:issue_id).where(repo_subscription_id: sub.id))
```
tiagofsilva: Instead of `IssueAssignment.select`, which will still generate AR objects, why not go with `pluck`, avoiding AR objects altogether?

app/models/user.rb
```diff
@@ -167,7 +167,7 @@ def days_since_last_clicked
   end
 
   def days_since_last_email
-    last_sent_at = repo_subscriptions.where("last_sent_at is not null").order(:last_sent_at).first.try(:last_sent_at)
+    last_sent_at = repo_subscriptions.where.not(last_sent_at: nil).order(:last_sent_at).first.try(:last_sent_at)
```
tiagofsilva: Nice! How about using `&.` instead of `try`? It's shorter, faster (even though not a problem here) and ruby standard.

app/models/issue_assigner.rb
anmolarora: Makes sense

app/models/issue_assigner.rb
anmolarora: That'll be better.

app/models/user.rb
anmolarora: Sure.

anmolarora:
@schneems Can this PR be merged?

schneems (commented):
I asked a question but apparently forgot to hit submit on it

app/models/issue_assigner.rb
```diff
@@ -19,28 +19,10 @@ def assign!
...
+                 .order(Arel.sql('RANDOM()'))
+                 .first
```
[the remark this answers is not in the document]
tiagofsilva: It does an extra trip to the DB for the subquery, but it does not instantiate any AR objects after that. Also it's a really quick subquery (only searches for id, which should be indexed). So it should be a little slower, but negligible IMO.

PR #953, opened 2019-04-02
Fix upgrade to Rails 5.2.3

prathamesh-sonpatki:


prathamesh-sonpatki:
cc @schneems The issue was with Rails 5.2.3 and will_paginate gem. It is fixed now.

Issue #963, opened 2019-04-14
Magic comment missing in all files


anmolarora:
https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/

PR #964, opened 2019-04-14
Added frozen string literal magic comment

anmolarora:
Related to #963


anmolarora:
> @anmolarora Can you add the rubocop cop for magic comment as well? And then run rubocop auto correct?

@prathamesh-sonpatki Done.

PR #975, opened 2019-04-26
add more tests for UsersController

GabrielRMuller:
added tests for `show`, `update` and `token_delete` in `users_controller`

### DIFF
--- a/test/functional/users_controller_test.rb
+++ b/test/functional/users_controller_test.rb
@@ -5,6 +5,34 @@ require 'test_helper'
 class UsersControllerTest < ActionController::TestCase
   include Devise::Test::ControllerHelpers
 
+  test '#show' do
+    sign_in users(:mockstar)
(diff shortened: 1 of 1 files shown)


test/functional/users_controller_test.rb
```diff
@@ -5,6 +5,34 @@
...
+    assert_redirected_to user
+    assert_equal 'New Name', user.name
+  end
+
+  test '#token_delete' do
+    sign_in users(:schneems)
+
+    get :token_delete, params: { account_delete_token: users(:schneems).account_delete_token }
+
+    assigns(:lonely_repos).any?
```
schneems, line 33: You are missing an assertion here, otherwise great tests. Thanks!

GabrielRMuller: indeed,  sorry about that, I was new to automated tests back then heheh!

Issue #954, opened 2019-04-04
Show the number of code helpers on repository list item.


nipe0324:
### Purpose

The number of code helpers are biased in spite of the importance of the repositories.
When the people who want to triage is exploring repositories, it is easer to choice better.

### Image

<img width="342" alt="スクリーンショット 2019-04-05 3 11 16" src="https://user-images.githubusercontent.com/2967009/55578360-a8a69380-5750-11e9-908b-85f538bafa1e.png">

PR #1082, opened 2019-09-17
Amount code helpers

edumoreira1506:
## Description

Hello guys! I saw the issue #954 that said about the amount of code helpers in a repository, so I think than add a new field on database should be complicated because have a lot of data and would be very heavy to update all of them.
So my solution is to use the GitHub API, in API we can get all contributors of each single repository, and the contributors are exactly the code helpers of the issue said about. 
Maybe we can change to other form, but i hope that helps :)

Image example:
Screenshot from 2019-09-18 10-10-19

### DIFF
--- a/app/models/repo.rb
+++ b/app/models/repo.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'docs_doctor/parsers/ruby/yard'
-require 'open-uri'
+require 'net/http'
 require 'json'
 
 class Repo < ActiveRecord::Base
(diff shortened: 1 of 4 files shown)


edumoreira1506:
Sorry, the image was of other PR, now is right!

app/models/repo.rb
```diff
@@ -108,6 +110,11 @@ def weight
     end
   end
 
+  def amount_code_helpers
+    apiResponse = Net::HTTP.get_response(URI.parse("https://api.github.com/repos/#{full_name}/contributors"))
```
YanCarlos: I think the model should not be responsible to implement an external request. Maybe you can use a service object or an interactor to implement it. What do you think?

PR #1157, opened 2019-10-26
Attempt to Reduce memory requirements of Doc Parse

schneems:
When parsing documentation we generate a lot of active record objects:

Most of them are only needed for a split second or two and can then be collected, however due to the QueryCache the objects will live on in memory for the entire duration of the job:

- https://github.com/rails/rails/issues/27002
- https://github.com/rails/rails/issues/28646

This PR disables the Query cache for the populate docs job. 

In the future we need to remove all these N+1 queries from the background doc parse, but for now hopefully this bandaid will at least prevent N+1 objects from being retained in memory for the duration of the job.

### DIFF
--- a/app/jobs/populate_docs_job.rb
+++ b/app/jobs/populate_docs_job.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 class PopulateDocsJob < RepoBasedJob
+  around_perform do |_job, block|
+    ActiveRecord::Base.uncached do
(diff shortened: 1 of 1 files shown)


schneems:
I think it helped a little but didn't keep the task from going over the memory limit and eventually doubling it:



I'm thinking that we need to manually add some logging of memory size for the YARD generated objects in addition to working on getting rid of the N+1 queries.

PR #1193, opened 2019-12-16
Trailing Spaces Removal

HassanAkbar:
Trailing Spaces Removal from .yml Files


prathamesh-sonpatki:
@HassanAkbar They might be added back when the VCR files are regenerated.

HassanAkbar:
@prathamesh-sonpatki Do you regenerate the test suite's HTTP interactions?
If you think that these test suite will be generated again then surely it does not makes sense to remove them.

prathamesh-sonpatki:
@HassanAkbar Yes, they might be re-generated. The VCR files should not be edited manually as they are auto-generated.

PR #1241, opened 2020-01-28
[Close #1145] Updated new repo URL parsing logic 

### LINKED ISSUE #1145: `https://www.codetriage.com/hexya-erp/hexya?files=1` should not be a valid repo name
Apparently, some URLs are still valid from GitHub's API perspective so they pass our exists check but they mess up future API requests one example is https://www.codetriage.com/hexya-erp/hexya?files=1.

We should disalow any characters in the name or user_name of a Repo that github would not allow i.e. `?` and `=`

ericc572:
Allow support for different types of URL including: 
```schneems/wicked
github.com/schneems/wicked
https://github.com/schneems/wicked"
https://github.com/schneems/wicked?files=1" # strip out everything after the ?```

### DIFF
--- a/app/controllers/repos_controller.rb
+++ b/app/controllers/repos_controller.rb
@@ -113,14 +113,13 @@ class ReposController < RepoBasedController
   def parse_params_for_repo_info
     return unless params[:url]
     params[:repo] ||= {}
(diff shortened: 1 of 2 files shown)


schneems:
Thanks!

Issue #1294, opened 2020-05-03
Language menu show all languages twice


AmrDeveloper:
In the Main layout when I opened the language menu I got all languages twice in the menu
I tested in from Desktop pc and android device

## Expected Behavior
The languages menu must show all languages once

## Current Behavior
The languages menu show all languages twice

## Screenshot
![screenshot](https://user-images.githubusercontent.com/23631699/80926024-c18ad980-8d94-11ea-8d12-d31e5b026ae1.PNG)

## Your Environment
* Browser Name and version: Firefox 75
* Operating System and version (desktop or mobile): Windows 10

ashishra0:
If this issue is sorted out, then close this issue 👍

PR #1297, opened 2020-05-07
Fix/covid language bug

edumoreira1506:
## Description
Hello Guys! I found a bug in new homepage. The bug was caused because the new area of covid. Now, are two projects filters, one just for covid related projects and another for the normal projects, so the both filters were synced, like you change a value from one input, the another will change too. Besides that when you open a filter, the another opens too.

## Related Issue
#1294 

## Motivation and Context
Fix the bug.

## Screenshots (if appropriate):
Before:
Peek 2020-05-07 16-51

After:
Peek 2020-05-07 16-50

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/app/assets/javascripts/homepage.js
+++ b/app/assets/javascripts/homepage.js
@@ -1,15 +1,15 @@
 // JS for the buttons on the home page
 $(document).ready(function(){
   $(".types-filter-button").click(function() {
-    $(".types-filter").toggle();
+    $(this).parent().find('.types-filter').toggle();
   });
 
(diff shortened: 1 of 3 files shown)


schneems:
Thanks!

Issue #1305, opened 2020-05-15
Incorrect query string parameter for language


swapnilmarkhedkar:
The query string  ```?language_covid=``` reflects no changes. However, ```?language=``` yeilds the right result, but it is not used.

## Expected Behavior

![Screenshot from 2020-05-16 01-46-17](https://user-images.githubusercontent.com/34151176/82093144-33114300-9718-11ea-9c3c-89241b526618.png)
![Screenshot from 2020-05-16 01-46-38](https://user-images.githubusercontent.com/34151176/82093161-3e646e80-9718-11ea-9841-df0d4cced811.png)
![Screenshot from 2020-05-16 02-05-15](https://user-images.githubusercontent.com/34151176/82093942-bc754500-9719-11ea-9e38-d2447e151117.png)

## Current Behavior

![Screenshot from 2020-05-16 01-48-24](https://user-images.githubusercontent.com/34151176/82093330-8b484500-9718-11ea-9e1c-d500e3403685.png)
![Screenshot from 2020-05-16 02-04-34](https://user-images.githubusercontent.com/34151176/82093936-ba12eb00-9719-11ea-88fb-5aa8ccd5813c.png)

Additionally, changes are reflected in only the selected dropdown field, however, it must be shown for both the dropdowns  fields.

## Possible Solution

A quick fix would be changing  query string  ```?language_covid=``` to  ```?language=```

## Steps to Reproduce (for bugs)

1. Select a language from the dropdown

## Context

I was trying to narrow down to a language.

## Your Environment

* Browser Name and version: Chrome Version 81.0.4044.138 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): Linux Mint 19.2 Cinnamon

## Additional request
It would be great if I could be assigned to work on this issue.

schneems:
> It would be great if I could be assigned to work on this issue.

That would be great. Take a stab at it! Thanks for reporting this

Cueball:
Came here to report what I think is the same issue, since I keep getting the email nags, but couldn't filter on projects where I might want to try to contribute.

Was going to add some more extended information on #1314 but then checked PRs and the PR above seems to solve an issue that's affected at least 2 people from filtering projects correctly [and consequently walking away until the next nag email arrives...come back, check, still broken, OK, try again later.]

That said, can say I don't understand the fix in #1307, but it *seems* to be an issue with a low-impact fix that's been known of for over a month now.

[...]

PR #1307, opened 2020-05-24
Fix filtering by lang with covid and without

LynxEyes:
## Description

Simple fix to the homepage language filters that is slightly broken since the introduction of the covid project section.

## Motivation and Context

On the homepage we currently have 2 language filters - one for covid project, other for normal stuff like CodeTriage always had.

Problem is that these 2 filters use slightly different querystring parameters but the view snippet that generates either of them is cached using the same cache key.
This results in both filters working on the same querystring parameter and given the order they exist on the page, right now it is only possible to filter by "language_covid".

## How Has This Been Tested?

By booting the app and clicking around 😭 

I've tried to add a test:

```ruby
  test "has filters for both covid and non-covid projects" do
    visit "/"

    assert page.has_selector?("a[data-query=language_covid]", text: "Ruby")
    assert page.has_selector?("a[data-query=language]", text: "Ruby")
  end
```

but seems like caching is always off on the test env.. which sounds odd given the config.. I need help here...

## Types of changes

Bug fix

## Checklist:

- [ ] ~My change requires a change to the documentation.~
- [ ] ~I have updated the documentation accordingly.~
- [ ] I have added tests to cover my changes.

### DIFF
--- a/app/views/pages/index.html.slim
+++ b/app/views/pages/index.html.slim
@@ -20,7 +20,7 @@
       = params["language_covid"] || "Select Language"
 
     ul.types-filter
-      = cache(:language_list, expires_in: 1.hour) do
+      = cache(:language_list_covid, expires_in: 1.hour) do
         - Repo.select("DISTINCT language").map(&:language).compact.sort.each do |lang|
           - unless lang.nil?
             li = link_to lang, "#", data: { toggle: "tab", language: lang, query: "language_covid" }


LynxEyes:
Please note that the test I added on the PR notes passes.. but also passes on unchanged code and that's why I didn't add it in.

I'd like to, I just don't know how to get caching to actually cache for that specific test... specially since `config/environments/test.rb` is set to cache to memory and that doesn't seem to happen.. IDK why..

swapnilmarkhedkar:
Hey looks like you are solving this issue raised by me https://github.com/codetriage/CodeTriage/issues/1305. So I'll close it when your PR gets accepted?

schneems:
Thanks!

PR #1310, opened 2020-05-27
Use Frontmatter gem to render universities

calebthompson:


calebthompson:
@schneems this is done now - open to feedback on the API for ActionPage here.

config/routes.rb
```diff
@@ -71,4 +64,7 @@
       end
     end
   end
+
+  get "/university", to: "universities#index", as: :university_index
+  get "/:id", to: "universities#show", as: :university
```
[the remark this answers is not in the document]
calebthompson: Squash and rebase were already separate files, and both of them still work. Or am I missing something?

app/controllers/university_controller.rb
```diff
@@ -1,39 +0,0 @@
...
-      @page_title = "Please Rebase your commits"
-      set_title(@page_title)
-      set_description("What on earth does rebasing commits mean and how do you do it? Find out the what, why, and how of git rebases here.")
-      render 'rebase'
-    when "reproduction_code", "reproduction", "reproduction_case"
-      @page_title = "Please Provide Reproduction Code"
-      set_title(@page_title)
-      set_description("Get answers for your open source bug reports faster by providing a code that reproduces your problem. Find out how.")
-      render 'reproduction_code'
-    when 'example_app', 'example_apps', 'example-app', 'example-apps'
```
calebthompson, line 27: This and the above reproduction lines have ids that no longer work. I don't have a good solution offhand except to keep hard-coding them in the controller.

db/schema.rb
```diff
@@ -27,8 +27,8 @@
     t.integer "repo_subscription_id"
     t.integer "doc_method_id"
     t.integer "doc_class_id"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
```
calebthompson, line 31: Oops.

calebthompson, line 31:
```suggestion
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
```

config/routes.rb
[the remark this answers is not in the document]
schneems: > Squash and rebase were already separate files, and both of them still work. 

Ahh, i'm confused by my own wrork

schneems:
> .fm

That would work it looks like it's currently only used for a weird adobe thing https://fileinfo.com/extension/fm .

I didn't know "Front matter" was a thing until you told me though. I think if you explain it in the readme, then it makes sense

calebthompson:
After giving it some more thought, I think .yaml is the best way forward because I do want to be able to support (.html.erb.).json and .toml in the future, and I don't think there's as good a way to differentiate those with file hints as there is with file extensions.

calebthompson:
Updated gem name to `frontmatter`

app/views/universities/show.html.slim
```diff
@@ -0,0 +1 @@
+= page_content(@page).html_safe
```
calebthompson, line 1: I'm currently plugging this into my site and realizing that I really dislike this way of rendering the pages.

I'm thinking that it makes more sense to do one of two options that are more Rails-y, and I'm interested in feedback on them:

## Layout template with direct page rendering

Have Frontmatter::PostsController:

1. Render the pages directly and assume that there's a template for pages instead of rendering into a show action:

   ```ruby
   def show
     # ...
     render "#{page_class.view_key}/#{@page.slug}", layout: "post"
   end
   ```

   ```erb
   <!-- a/v/layouts/post.html.erb -->
   <article class="post">
    <%= yield %>
   </article>
   ```

   ```markdown
   <!-- a/v/posts/my-post.md.yaml -->
   ---
   title: My Post
   date: 2020-01-01
   ---

   My first post
   ```
2. Provide the render bit in a controller-level helper method for overriding:

   ```ruby
   class PostsController < Frontmatter::PostsController
     def show
       # ...
       render_page
     end
   end
   ```

This method has the advantage that it uses familiar features such as layouts, with the drawback that it's surprising not to see a show view template for a show action.

## show view file with partial pages

Rather than using an application-level template (though it needn't be used elsewhere of course), use the show view file and render pages as partials. This necessitates that all pages be named `_<slug><extension>[other extensions]<.[yaml][json][toml]>` such as `_squash.md.erb.yaml`.

```erb
<!-- a/v/pages/show.html.erb -->
<article class="page show">
  <%= render @page %>
</article>
```

In this case, Frontmatter::Page would provide a to_partial_path of `<tableize>/<slug>` such as "universities/squash" for an object `#<University slug="squash">`

The advantage that the view files are all within a single directory and that the API is sensible without requiring new helpers is nice, I think. The drawbacks are that it requires renaming files with the leading underscore and that it overrides to_partial_path which could be surprising.

Do you have any thoughts on those as alternate APIs to the `page_content(@page).html_safe` helper? As it stands the API is a little closer to the "show view file with partial pages" option, but with additional API in the form of the helper.

calebthompson: Because of template inheritance, Frontmatter can also provide a default, empty template or show page so that those pages don't need to exist if they're not wanted in an app.

PR #1396, opened 2020-11-15
Rails 6.1.0.rc1

prathamesh-sonpatki:
## Description
Update to Rails 6.1.0.rc1


schneems:
thanks! It looks like you didn't run into this issue with missing rake tasks https://github.com/heroku/heroku-buildpack-ruby/issues/1099

prathamesh-sonpatki:
No didn't face it locally. Will check in detail

schneems:
I think the reason it didn't affect us in our test environment is that in addition to loading from schema we also run `rake db:migrate` and CodeTriage's migrations are clean enough that we can migrate from scratch while all apps do not have this ability.

Issue #1406, opened 2020-12-11
Source Locations for Skylight


gitKrystan:
Thank you so much for participating in the [Skylight for Open Source](https://www.skylight.io/oss) program. We'd like to invite CodeTriage to participate in the public alpha for Source Locations for Skylight!

With this new feature, Skylight can help you pinpoint the locations in your code that correspond to events in the Skylight Event Sequence. This feature will allow your contributors to more easily find out exactly where an expensive operation originated without having to scour your code.

![https://d2ryfneqiwuan6.cloudfront.net/assets/skylight/docs/features/source-locations-popover-1181b5debba553a2b06bd914405997dce0334388b69fe3877c30acc21393ca25.png](https://d2ryfneqiwuan6.cloudfront.net/assets/skylight/docs/features/source-locations-popover-1181b5debba553a2b06bd914405997dce0334388b69fe3877c30acc21393ca25.png)

Source locations are available with the beta2 version of the gem you are currently running, but there are some improvements in this latest version that would be useful to have before we enable the flag. (See [CHANGELOG](https://github.com/skylightio/skylight-ruby/blob/master/CHANGELOG.md#500beta4)). I can open a PR to upgrade you to beta4 if you'd like.

You can find documentation about the new feature at [https://www.skylight.io/support/source-locations](https://www.skylight.io/support/source-locations), and as always, if you have any questions or feedback, please get in touch with us either by responding here, emailing [support@skylight.io](mailto:support@skylight.io), or via the in-app messenger at the lower-right of your Skylight dashboard.

Let us know if you are interested, and we will enable the feature flag for you on Skylight's servers.

PR #1407, opened 2020-12-11
Upgrade Skylight gem to 5.0beta4

gitKrystan:
## Description
Upgrade Skylight gem to 5.0.0.beta4

## Related Issue
https://github.com/codetriage/CodeTriage/issues/1406

## Motivation and Context
With 5.0beta4, we can enable the Skylight for Source Locations feature for your account. This feature will allow your contributors to more easily find out exactly where an expensive operation originated without having to scour your code.

## How Has This Been Tested?
Existing tests pass locally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,7 +110,7 @@ gem 'redis-namespace'
 gem 'stackprof'
 gem 'flamegraph'
 gem 'prawn'
-gem 'skylight', '5.0.0.beta2'
+gem 'skylight', '~> 5.0.0.beta4'
 
 gem 'minitest', '5.14.1'
 gem 'sitemap_generator'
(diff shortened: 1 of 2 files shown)


gitKrystan:
Thanks for merging! I've enabled the feature flag on our end. Because the feature is still in alpha, it is not yet available on your public Skylight for OSS dashboard. While logged in, however, the app owner and any collaborators should be able to see the feature (once you deploy the new agent).

You might also consider enabling the [deploy tracking feature](https://www.skylight.io/support/advanced-setup#deploy-tracking) if it's not already enabled. Source Locations works best if you have enabled deploy tracking, because it allows you find an event’s source location for a particular moment in time and track changes across deploys.

Issue #1686, opened 2022-10-05
Dependency parser to find repos


schneems:
## Problem

People don’t know what repos to subscribe to. My best suggestion is to look in their dependencies (such as Gemfile.lock, package.json, or Pom.xml)

For now I’m asking them to manuall look in their dependencies and then search for repos on CodeTriage. I would love to automate this process.

I’m thinking it looks like this: the user uploads a file then based on the name we run an extraction script that returns the dependencies in the file in something machine readable (outputting it in json to STDOUT and any errors or warnings to STDERR).

```
$ cat Gemfile.lock | parse_gemfile.rb
{ 'repos': [ { 'name': "activerecord" } ], 'language': "ruby" }
```

Then we can take that information to search of a project with that language and name exists in the database.

## Ruby tooling - Task zero

We can use the lockfile parser that ships with Ruby to do this for Ruby. See the heroku/heroku-buildpack-ruby for some inspiration. 

- [ ] Ruby (Gemfile.lock)

In addition to doing this code I also would like a test file setup where we can put future parser tests. That will make it easier for future PRs to be tested (if they just have to copy an existing test instead of learning Ruby/Rspec if they don't know it.

Place the test in `test/dependency_parsers/ruby_parser_test.rb`

I still want the code to be implemented as a script. Put it in a folder under `lib/dependency_parser/ruby/parse.rb`. The input to the script will be the file contents via STDIN. I want json to STDOUT and errors to STDERR. Failures should result in non-zero exit code

## Dependency/lockfile parsing for other languages- Task one

Ideally we won’t need to install other tooling, for example installing maven would be overkill. It takes a bunch of time to install and eats up a bunch of space on disk. This would slow down deploys.

I'm thinking we need to extract the information in either Ruby or Bash or Javascript as these are the three main languages already on the app.

We need a parser for these languages:

- [ ] Javascript (package.json)
- [ ] Java 
- [ ] Python
- [ ] Golang
- [ ] Rust
- [ ] PHP
- [ ] Swift
- [ ] Elixir
- [ ] Your favorite language I didn't mention here.

Each of these can be in a different PR. Put them in a folder under `lib/` named `dependency_parser/<language>/parse.<extension>`. When sending a PR please give me an example input and expected output. 

The script needs to defensively check for invalid input and output a helpful error message if:

- It's formatted incorrectly
- Is missing critical information
- Is empty

If a script cannot continue then it needs to exit with non-zero status code.

[...]

PR #1689, opened 2022-10-05
Add Gemfile.lock parsing for dependency parser

khasinski:
Dependency parsing for #1686

Uses Bundler gem for parsing, parses only name for now, we should probably extract URL for repos.

## Related Issue
- #1686 Gemfile.lock, more languages to come, stay tuned ;) 

## Motivation and Context
An initial stub to get the dependency parsing running

## How Has This Been Tested?
- invalid file + a tiny Gemfile.lock, might need more testing

## Checklist:

### DIFF
--- a/lib/dependency_parser/ruby/parse.rb
+++ b/lib/dependency_parser/ruby/parse.rb
@@ -5,26 +5,58 @@ require 'json'
 module DependencyParser
   module Ruby
     class Parse
+      attr_reader :errors
+
       def initialize(content)
         @content = content
+        @errors = []
+        @direct = []
       end
 
       def call
-        deps = Bundler::LockfileParser.new(content)
-        parsed = deps.specs.map do |spec|
-          { repos: [{ name: spec.name }], language: "ruby" }
-        end
-        JSON.pretty_generate(parsed)
+        validate_lockfile!(content)
+        return if error?
+
+        @direct = Bundler::LockfileParser.new(content).specs.map do |spec|
+          fetch_spec(name: spec.name, version: spec.version)
+        end.compact
+      end
+
+      def direct
+        { repos: @direct, language: "ruby" }
+      end
+
+      def success?
+        errors.empty?
+      end
+
+      def error?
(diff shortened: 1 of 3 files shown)


lib/dependency_parser/ruby/parse.rb
```diff
@@ -0,0 +1,50 @@
...
+        { repos: @direct, language: "ruby" }
+      end
+
+      private
+
+      attr_reader :content
+
+      def self.valid_gemfile?(content)
+        Bundler::LockfileParser.new(content).specs.any?
+      rescue
```
schneems, line 29: I would like to capture errors and make them accessible later.

lib/dependency_parser/ruby/parse.rb
```diff
@@ -0,0 +1,50 @@
...
+        false
+      end
+
+      def fetch_spec(name:, version: nil)
+        full_spec = fetcher.fetch_spec([name, version])
+        { name: full_spec.name, url: extract_url(full_spec), description: full_spec.description }
+      rescue
```
schneems, line 36: Was there a class of error that triggered this specific rescue? I'm just curious. The other rescue looks like a "passed in the wrong file" kind of thing. I feel like we should capture this error too, but I'm not entirely sure what the failure modes that would trigger the rescue here are.

schneems (changes requested):
Awesome work. I asked a few questions and asked for some fairly minor changes. Slack me if you wanna chat about any of it.

schneems:
Any updates? This was pretty close to over the line.

khasinski:
Hey, I'll finish it up today

lib/dependency_parser/ruby/parse.rb
khasinski: I haven't found any. I can remove it, but it seems like it's better to log the failed spec instead of failing the entire file, we can at least get some log from it.

schneems (approved):
Thanks!

Issue #1735, opened 2022-10-17
Remove aseem-thakar/ozbargain-plus from CodeTriage


aseem-thakar:
Hi, this repository no longer exists, has no subscribers but is still listed on CodeTriage
Can it please be removed?

https://www.codetriage.com/aseem-thakar/ozbargain-plus

Issue #1704, opened 2022-10-08
Remove archived repos


hbontempo-br:
Last week I received I "CodeTriage misses you" email (yeah, I know... 😞) that suggested me to engage with an archived repo.
<img src="https://user-images.githubusercontent.com/15916283/194674617-554b0870-fac4-409b-a649-b059afd4f4d1.png" width="300">

I don't think that any archived repository should be suggested to a user (both on email suggestions and through the web application)

## Expected Behavior

Archived repos probably should be removed from the platform. 
They shouldn't be suggested on daily emails or through the web interface "Open source projects on GitHub that need your help." functionatlity

## Current Behavior

Archived repos are handled the same way as any other repo.

## Possible Solution

Periodically check if the repo was archived then:
- Record it:
  - Save to a flag column  in the database
  - Add a [default_scope](https://apidock.com/rails/v6.1.3.1/ActiveRecord/Scoping/Default/ClassMethods/default_scope) to the model

OR

- Delete it

(Probably could expand the the existing [Checks if repos have been deleted on GitHub](https://github.com/codetriage/CodeTriage/blob/cfabdf6e9417255ee1e077d355861460d351d5a9/lib/tasks/schedule.rake#L41) task if necessary to unnecessary requests to GH's api)

## Steps to Reproduce (for bugs)

Example: https://www.codetriage.com/simplabs/excellent

schneems:
100% we should remove archived repos and send an email to subscribers.

And with deleted repos. And we should support moved/transferred repos as
well. Depending on complexity this could be several PRs.

Thanks for the issue!
-- 
Richard Schneeman
https://www.schneems.com
he/him

hbontempo-br:
Didn't though about about this other scenarios.

I didn't found anything explicitly on GH's documentation on what to expect when a repo is rename or transferred, so I did some testing with my accounts.
Renaming and transferring the repo private shouldn't be a bib problem to track, despite all the changes the repo can still be accessed using the repo endpoint with the old address (unless the a new repo is created with the name of old repo).
Making the repo private results in the same behaviour as a delete repo (response 404).

avcwisesa:
I also encountered a slight doubt when using CodeTriage the first time related to moved/transferred repo not sure which one is the correct one.   

<img width="638" alt="image" src="https://user-images.githubusercontent.com/18155327/194808553-4139d68d-fc0f-4e45-9089-ca5af9136168.png">

[...]

Issue #1545, opened 2021-08-30
Remove repo I submit earlier


ShmuelTreiger:
I added the Couchers project, we're looking to remove it

https://www.codetriage.com/couchers-org/couchers

Looked around, didn't see any easy way to remove it

Parimalisgreat:
@ShmuelTreiger , what do you mean by removing the project, do you want to remove it from frontend?

ShmuelTreiger:
Yes. From codetriage.com so that it doesn't show up anymore

schneems:
Are you able to unsubscribe from the project or the UI won’t let you?

ShmuelTreiger:
I am not looking to unsubscribe. I want couchers.org removed from codetriage.com. For example, when I search for projects on codetriage, I don't want to find the couchers.org project. I don't know how else to describe it

ShmuelTreiger:
I'm not asking for a feature, I'm just asking for my project to please be removed.

More to the point, this feature was requested in 2016 and still isn't available. I'd like to have my project removed before 2026.

ShmuelTreiger:
1) I added it. I suspect it's unlikely someone else adds it back. I don't know the other person subscribed, but I doubt they would add it back.

2) I would suggest having a blacklist, and allowing projects to permanently remove themselves instead of allowing anyone to merely add it back once it's removed.

3) We want to pivot away from sort of "green" devs toward people with more experience. I added us to codetriage as well as a few similar projects, and we found that the devs it was bringing in would come just long enough for us to get them up to speed and then disappear (classic problem, but we're all unpaid volunteers and don't have the time for it).

ShmuelTreiger:
We do a pretty good job of recruiting active users, we think. It's a social-media adjacent website, so our members visit the website somewhat regularly.

>I'm emailing that user to ask them if I can remove them from the subscription. If they say yes then I'll delete it from the site.

Appreciate it, thanks.

>If you ever figure out that problem, let me know. It's hard to find the right balance for investing in community building when you don't know if that community will re-engage in the future.

Will do. It's definitely a challenge.

Holeryn:
I have a similar "problem", i archived my repo because i find it completely useless: https://www.codetriage.com/holeryn/asm_utils
P.S: no problem with the site, it's great but i don't even remember why i asked help for a such a stupid thing.

PR #1743, opened 2022-10-23
Changes default behaviour to ignore archived and deleted repos

hbontempo-br:
## Description

Changes default behavior of ReposModel to deleted and archived repos.
This repos are going to remain on database but should be ignored everywhere in the application

## Related Issue

Related to #1704
Should deal with #1735 / #1545

## Motivation and Context

Should prevent "bad" repos to be presented to users

## How Has This Been Tested?

Automated tests + rubocop analysis (locally)

## Screenshots (if appropriate):
--

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


app/models/repo.rb
```diff
@@ -24,6 +24,9 @@ class Repo < ActiveRecord::Base
 
   CLASS_FOR_DOC_LANGUAGE = { "ruby" => DocsDoctor::Parsers::Ruby::Yard }
 
+  default_scope -> { where(removed_from_github: false) }
+  default_scope -> { where(archived: false) }
```
schneems, line 28: I'm strongly opposed to `default_scope` but enumerating why is difficult. It's just one of those things that I've built up scar tissue from over the years. It saves a few minutes now and really hurts later when debugging etc. 

I'm asking on twitter for alternatives. So far I've got:

- Make it a regular scope and call it everywhere. (not so different than `with_some_issues` which is used heavily https://github.com/codetriage/CodeTriage/blob/aa03ffcd56f57c2b12ffb48a752ae27ec3ea93b7/app/models/repo.rb#L176
- Split up logic into different classes. So you might have a AllRepo  class and an ArchivedRepo class etc.  I'm not sure what we would call the class that ignores deleted and archived, maybe `ActiveRepo` class ? 
- Might get more replies on twitter

I think I would lean towards making it a regular scope and applying it as needed.

schneems (changes requested):
Please change out `default_scope` for something else.

app/models/repo.rb
hbontempo-br: Thanks for the feedback!
I don't have much experience, will take a look in the tweeter responses and search a little bit.
I will come back soon with a alternative.

hbontempo-br: I read a bit and `default_scope` is really broadly hated in the community.
Changed to a 'normal' scope, it's the simplest alternative.

hbontempo-br:
PR ended up being bigger than originally expected.
Could break up into smaller ones if you prefer.

schneems (approved):

schneems:
Thanks for working on this. Sorry for the EXTREME delay. Looks great. I corrected a merge conflict. re-running tests now.

